PR #12, opened 2019-07-30
Add rocksdb

Stumble:
https://github.com/tendermint/tendermint/issues/3831

Using my fork of gorocksdb from https://github.com/tecbot/gorocksdb for two reasons: 1. The upstream repo encourages vendoring for API stability. 2. The default LDFLAGS setting in upstream repo enforces linking unnecessary compression libraries(https://github.com/tecbot/gorocksdb/blob/162552197222a834c6857385ae890e4decdd5049/dynflag.go#L5), so that if the library is not installed, link will fail. IMHO, tendermint should fork this wrapper and use it here.

To run tests:
First, install rocksdb,
```
homedir="$PWD"
sudo apt-get update
sudo apt-get install -y build-essential gcc g++ make libgflags-dev libsnappy-dev
wget -O rocks.tar.gz https://github.com/facebook/rocksdb/archive/v6.1.2.tar.gz && \
  tar -zxvf rocks.tar.gz && \
  cd rocksdb-6.1.2/ && \
  make static_lib -j4 && \
  sudo cp librocksdb.a /usr/local/lib/ && \
  cd include/ && \
  sudo rm -rf /usr/local/include/rocksdb && \
  sudo cp -r rocksdb /usr/local/include/ && \
  sudo ldconfig
cd "$homedir"
rm rocks.tar.gz
rm -rf rocksdb-6.1.2
```
Run tests:
```
GO_LDFLAGS="-lsnappy" go test -v -tags "rocksdb"
```


db/db.go
```diff
@@ -32,6 +32,11 @@ const (
 	//   - may be faster is some use-cases (random reads - indexer)
 	//   - use boltdb build tag (go build -tags boltdb)
 	BoltDBBackend DBBackendType = "boltdb"
+	// RocksDBBackend represents rocksdb (uses tecbot's wrapper)
+	//   - EXPERIMENTAL
+	//   - requires gcc
+	//   - use boltdb build tag (go build -tags rocksdb)
```
melekes:
```suggestion
	//   - use rocksdb build tag (go build -tags rocksdb)
```

db/c_level_db_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
-	cmn "github.com/tendermint/tm-cmn/libs/common"
+	cmn "github.com/tendermint/tm-cmn/common"
```
melekes: this got changed I think. Sorry, we're in a process of separating some libraries.

db/rocks_db.go
```diff
@@ -0,0 +1,342 @@
...
+func NewRocksDBWithOptions(name string, dir string, opts *gorocksdb.Options) (*RocksDB, error) {
+	dbPath := filepath.Join(dir, name+".db")
+	db, err := gorocksdb.OpenDb(opts, dbPath)
+	if err != nil {
+		return nil, err
+	}
+	ro := gorocksdb.NewDefaultReadOptions()
+	wo := gorocksdb.NewDefaultWriteOptions()
+	woSync := gorocksdb.NewDefaultWriteOptions()
+	woSync.SetSync(true)
```
melekes: what this line does?

db/rocks_db.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	return database, nil
+}
+
+// Implements DB.
+func (db *RocksDB) Get(key []byte) []byte {
+	key = nonNilBytes(key)
+	res, err := db.db.Get(db.ro, key)
+	if err != nil {
+		res.Free()
```
melekes: why not put this above with `defer`? that way we'll avoid duplication s.Free() in moveSliceToBytes and here.

Stumble: turns out `res.Free()` when `err != nil` is not necessary, removed. I would prefer to have a function that moves c bytes to go slice.

db/rocks_db.go
Stumble: Do you mean these variables? read/write options?

melekes: I mean `woSync.SetSync(true)`

melekes:
> Using my fork of gorocksdb from https://github.com/tecbot/gorocksdb for two reasons:

Have you tried to contribute back to original lib?

db/rocks_db.go
Stumble: `woSync.SetSync(true)` prepares a sync write option. https://godoc.org/github.com/tecbot/gorocksdb#WriteOptions.SetSync

Stumble:
> > Using my fork of gorocksdb from https://github.com/tecbot/gorocksdb for two reasons:
> 
> Have you tried to contribute back to original lib?

Nope, as vendoring was suggested for gorocksdb anyway. I just sent an issue in the upstream, let's see how it goes.

Stumble:
@melekes , the upstream has not respond to the issue yet. https://github.com/tecbot/gorocksdb/issues/167

Stumble:
Rocksdb's stats does not work with https://github.com/cosmos/cosmos-sdk/blob/e4c8bd72b72b18035ede103cb4a47da5825d330e/server/export.go#L95.

db.go
```diff
@@ -32,6 +32,11 @@ const (
 	//   - may be faster is some use-cases (random reads - indexer)
 	//   - use boltdb build tag (go build -tags boltdb)
 	BoltDBBackend DBBackendType = "boltdb"
+	// RocksDBBackend represents rocksdb (uses tecbot's wrapper)
```
melekes:
```suggestion
	// RocksDBBackend represents rocksdb (uses Stumble fork github.com/stumble/gorocksdb of the tecbot wrapper)
```

marbar3778:
@Stumble it seems gorocksdb has picked up steam could you rebase your fork until your PR get merged?

Stumble:
@marbar3778 The PR is merged, I suggest we switch to the original branch.

PR #14, opened 2019-07-31
Marko/tm cmn to db

marbar3778:


db/remotedb/grpcdb/client.go
```diff
@@ -4,7 +4,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 
-	protodb "github.com/tendermint/tm-cmn/db/remotedb/proto"
+	protodb "github.com/tendermint/tm-db/db/remotedb/proto"
```
melekes: Can we get rid of `db` directory and move its content into the root dir?

util.go
```diff
@@ -43,3 +45,36 @@ func IsKeyInDomain(key, start, end []byte) bool {
...
+	_, err := os.Stat(filePath)
+	return !os.IsNotExist(err)
+}
+
+const (
+	strChars = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz" // 62 characters
+)
+
+// Str constructs a random alphanumeric string of given length.
+func RandStr(length int) string {
```
melekes: Do we use this in tests? If yes, then we should unexport it and move into `test_helpers.go`

util.go
```diff
@@ -43,3 +45,36 @@ func IsKeyInDomain(key, start, end []byte) bool {
 	}
 	return true
 }
+
+func FileExists(filePath string) bool {
```
melekes: same here. should probably be unexported

marbar3778: used in fsdb.go

util.go
marbar3778: will move this to test_helpers.go

PR #19, opened 2019-09-19
Readme update

marbar3778:


melekes:
NOTE: We should write a proper readme.

PR #30, opened 2019-11-12
Return Errors instead of Panicing

### LINKED ISSUE #4: Change DB interface to return errors
Low-level libs (db) should not decide whenever to panic or not. It's better to let dbs return an error and handle it in the calling code (say retry N times and maybe panic afterwards).

Pros:
- standard interface
- error handling is the same for every database (memdb, go_level_db, ...) => ease of customization in the future

Cons:
- we'll need to modify calling code

### LINKED ISSUE #28: Do not panic - return error instead
Hi,

Can db methods be changed so they return error instead of panic. It will help a lot in handling error case without repeting `recovery` everywhere.

marbar3778:
- closes #28
- closes #4

Return errors and have the application decide how they want to handle panicing :smile


backend_test.go
```diff
@@ -76,10 +88,15 @@ func TestBackendsNilKeys(t *testing.T) {
 				// Nil keys are treated as the empty key for most operations.
 				expect := func(key, value []byte) {
 					if len(key) == 0 { // nil or empty
-						assert.Equal(t, db.Get(nil), db.Get([]byte("")))
+						nilValue, err := db.Get(nil)
+						assert.NoError(t, err)
+						byteValue, err := db.Get([]byte(""))
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

prefix_db.go
```diff
@@ -193,8 +201,8 @@ func (pdb *PrefixDB) Print() {
 	itr := pdb.Iterator(nil, nil)
 	defer itr.Close()
 	for ; itr.Valid(); itr.Next() {
-		key := itr.Key()
-		value := itr.Value()
+		key, _ := itr.Key()
```
golangcibot: Error return value of `itr.Key` is not checked (from `errcheck`)

prefix_db.go
```diff
@@ -193,8 +201,8 @@ func (pdb *PrefixDB) Print() {
...
+		value, _ := itr.Value()
```
golangcibot: Error return value of `itr.Value` is not checked (from `errcheck`)

prefix_db.go
```diff
@@ -267,7 +275,11 @@ type prefixIterator struct {
 }
 
 func newPrefixIterator(prefix, start, end []byte, source Iterator) *prefixIterator {
-	if !source.Valid() || !bytes.HasPrefix(source.Key(), prefix) {
+	key, err := source.Key()
+	if err != nil {
```
golangcibot: SA9003: empty branch (from `staticcheck`)

fsdb.go
```diff
@@ -43,19 +43,19 @@ func NewFSDB(dir string) *FSDB {
 	return database
 }
 
-func (db *FSDB) Get(key []byte) []byte {
```
marbar3778: Do we want to keep this DB around, I cant see a reason someone would want to use it?

melekes: For testing maybe?

prefix_db.go
```diff
@@ -267,7 +286,8 @@ type prefixIterator struct {
 }
 
 func newPrefixIterator(prefix, start, end []byte, source Iterator) *prefixIterator {
-	if !source.Valid() || !bytes.HasPrefix(source.Key(), prefix) {
+	key, _ := source.Key() //nolint:errcheck
```
marbar3778: reason for `nolint:errcheck` here is that it does come with an error quite often, even through the tests. 

Either I can ignore it like above, I can have this function return the error, or just print it out? Leave your opinion below while reviewing

boltdb.go
```diff
@@ -62,60 +62,66 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
...
-		panic(err)
+		return nil, err
 	}
 	return
 }
 
 func (bdb *BoltDB) Has(key []byte) bool {
-	return bdb.Get(key) != nil
+	bytes, err := bdb.Get(key)
+	if err != nil {
```
alessio: This may be problematic cause it hides errors. In other words, this function should not IMHO just return false if an error is raised, it really should panic

marbar3778: I was thinking about this as well, could return the error and let the application handle it, I want to avoid panicking in this case. I can switch out the `bool` for an `error` didn't know how this would be accepted as the name `Has()` to me sounds like it should return true or false, but if people are up for it returning the error I will make the change

backend_test.go
```diff
@@ -76,70 +94,95 @@ func TestBackendsNilKeys(t *testing.T) {
...
-				db.Delete(nil)
+				err = db.Delete(nil)
+				require.NoError(t, err)
 				expect(nil, nil)
 
 				// Set nil, Delete empty
-				db.Set(nil, []byte("abc"))
+				err = db.Set(nil, []byte("abc"))
+				require.NoError(t, err)
 				expect(nil, []byte("abc"))
```
jackzampolin: Do we still need these `expect`s? I think those were there to catch the panic no?

boltdb.go
```diff
@@ -62,60 +62,66 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
...
-func (bdb *BoltDB) Get(key []byte) (value []byte) {
+func (bdb *BoltDB) Get(key []byte) (value []byte, err error) {
 	key = nonEmptyKey(nonNilBytes(key))
-	err := bdb.db.View(func(tx *bbolt.Tx) error {
+	err = bdb.db.View(func(tx *bbolt.Tx) error {
 		b := tx.Bucket(bucket)
 		if v := b.Get(key); v != nil {
 			value = append([]byte{}, v...)
 		}
 		return nil
```
jackzampolin:
```suggestion
		return value, nil
```

boltdb.go
jackzampolin: I think in keeping with the rest of the functions having the signature be:
```go
func (bdb *BoltDB) Has(key []byte) (b bool, err error) {}
```

boltdb.go
```diff
@@ -62,60 +62,66 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
...
 	if err != nil {
-		panic(err)
+		return err
 	}
+	return nil
 }
 
-func (bdb *BoltDB) SetSync(key, value []byte) {
-	bdb.Set(key, value)
+func (bdb *BoltDB) SetSync(key, value []byte) error {
```
jackzampolin: What is the purpose of this function?

boltdb.go
```diff
@@ -62,60 +62,66 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
...
-		panic(err)
+		return err
 	}
+	return nil
 }
 
-func (bdb *BoltDB) DeleteSync(key []byte) {
-	bdb.Delete(key)
+func (bdb *BoltDB) DeleteSync(key []byte) error {
+	return bdb.Delete(key)
```
jackzampolin: Same here, this just looks like an alias for `Delete`

boltdb.go
alessio: IMHO @jackzampolin suggestion is the way to go

boltdb.go
marbar3778: in boltdbs case, yes but in other dbs (goleveldb) it does have different functionality. so to adhere to the interface it is there. We can speak about re working the interface, but not sure people will be up for that

boltdb.go
marbar3778: sounds good

boltdb.go
jackzampolin: Thank you!

boltdb.go
jackzampolin: Answered above.

prefix_db.go
alexanderbez: Why does it return an error quite often? We should figure out why and maybe alleviate such cases, or at least reduce the probability. Are these errors benign? I.e. if an error happens, will iteration still be valid? If so, I would simply ignore or add a debug statement. I'm OK with either. If it's not benign, then we need to fix the root cause.

prefix_db.go
```diff
@@ -293,29 +318,45 @@ func (itr *prefixIterator) Valid() bool {
...
+		return nil
 	}
+	return nil
 }
 
-func (itr *prefixIterator) Key() (key []byte) {
+func (itr *prefixIterator) Key() (key []byte, err error) {
 	if !itr.valid {
-		panic("prefixIterator invalid, cannot call Key()")
+		return nil, errors.New("prefixIterator invalid, cannot call Key()")
```
alexanderbez:
```suggestion
		return nil, errors.New("prefixIterator invalid; cannot call Key()")
```

alexanderbez: Above applies for all errors like such.

prefix_db.go
```diff
@@ -293,29 +318,45 @@ func (itr *prefixIterator) Valid() bool {
...
 	}
-	return stripPrefix(itr.source.Key(), itr.prefix)
+	return stripPrefix(key, itr.prefix), nil
 }
 
-func (itr *prefixIterator) Value() (value []byte) {
+func (itr *prefixIterator) Value() (value []byte, err error) {
 	if !itr.valid {
-		panic("prefixIterator invalid, cannot call Value()")
+		return nil, errors.New("prefixIterator invalid, cannot call Value()")
```
alexanderbez:
```suggestion
		return nil, errors.New("prefixIterator invalid; cannot call Value()")
```

.gitignore
```diff
@@ -11,5 +11,5 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
-/.vscode
+.vscode
```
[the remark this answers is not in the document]
melekes: https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore

prefix_db.go
marbar3778: according to the fucntion it was treated previously as ignored and checked in `!bytes.HasPrefix(source.Key(), prefix)` so the error is benign.

this is what I am interpreting as the check if the prefix is present. It may be easier to see if i clean it up. 
```
	if !source.Valid() || !bytes.HasPrefix(key, prefix) {
		return &prefixIterator{
			prefix: prefix,
			start:  start,
			end:    end,
			source: source,
			valid:  false,
		}
	}
```

fsdb.go
marbar3778: we can keep it around for now, but should look into cleaning up things that are incomplete(meaning panic("not implemented) is used in places ) or not being used

c_level_db.go
```diff
@@ -50,79 +51,95 @@ func NewCLevelDB(name string, dir string) (*CLevelDB, error) {
...
+	return nil
 }
 
 func (db *CLevelDB) DB() *levigo.DB {
 	return db.db
 }
 
 // Implements DB.
-func (db *CLevelDB) Close() {
+func (db *CLevelDB) Close() error {
```
melekes: not sure if returning an error here would benefit anybody really

go_level_db.go
```diff
@@ -41,73 +41,83 @@ func NewGoLevelDBWithOpts(name string, dir string, o *opt.Options) (*GoLevelDB,
...
 }
 
 // Implements DB.
-func (db *GoLevelDB) Has(key []byte) bool {
-	return db.Get(key) != nil
+func (db *GoLevelDB) Has(key []byte) (bool, error) {
+	if _, err := db.Get(key); err != nil {
+		return false, err
+	}
+	return true, nil
```
melekes: shouldn't we check that value != nil before returning true?

prefix_db.go
```diff
@@ -179,24 +192,35 @@ func (pdb *PrefixDB) DeleteNoLockSync(key []byte) {
...
-		key := itr.Key()
-		value := itr.Value()
+	var err error
+	for ; itr.Valid(); err = itr.Next() {
+		if err != nil {
+			return err
+		}
+		key, err := itr.Key()
+		if err != nil {
+			return err
```
melekes:
```suggestion
			return errors.Wrap(err, "key")
```

prefix_db.go
```diff
@@ -179,24 +192,35 @@ func (pdb *PrefixDB) DeleteNoLockSync(key []byte) {
...
+		}
+		value, err := itr.Value()
+		if err != nil {
+			return err
```
melekes:
```suggestion
			return errors.Wrap(err, "value")
```

prefix_db.go
```diff
@@ -179,24 +192,35 @@ func (pdb *PrefixDB) DeleteNoLockSync(key []byte) {
...
 
 	itr := pdb.Iterator(nil, nil)
 	defer itr.Close()
-	for ; itr.Valid(); itr.Next() {
-		key := itr.Key()
-		value := itr.Value()
+	var err error
+	for ; itr.Valid(); err = itr.Next() {
+		if err != nil {
+			return err
```
melekes:
```suggestion
			return errors.Wrap(err, "next")
```

rocks_db.go
```diff
@@ -64,55 +65,63 @@ func NewRocksDBWithOptions(name string, dir string, opts *gorocksdb.Options) (*R
...
 
 // Implements DB.
-func (db *RocksDB) Has(key []byte) bool {
-	return db.Get(key) != nil
+func (db *RocksDB) Has(key []byte) (bool, error) {
+		_, err := db.Get(key)
+		if err != nil {
+			return false, err
+		}
+		return true, nil
```
melekes: same.. we should check the value here

c_level_db.go
marbar3778: I was thinking the same but on some impl the close returns an error, boltdb close func returns an error so I didnt want to hide it

types.go
```diff
@@ -3,27 +3,27 @@ package db
...
 	// Delete deletes the key.
 	// A nil key is interpreted as an empty byteslice.
 	// CONTRACT: key readonly []byte
-	Delete([]byte)
-	DeleteSync([]byte)
+	Delete([]byte) error
+	DeleteSync([]byte) error
 
 	// Iterate over a domain of keys in ascending order. End is exclusive.
 	// Start must be less than end, or the Iterator is invalid.
```
marbar3778: Do we want to return errors on the iterators as well?

boltdb_test.go
```diff
@@ -20,6 +20,8 @@ func TestBoltDBNewBoltDB(t *testing.T) {
 	db.Close()
 }
 
+//TODO: ADD TESTS!!!
```
alexanderbez: Is this still a TODO?

krhubert:
LGTM - Great Job!

boltdb_test.go
marbar3778: yes, there is an issue for it. To keep the scope of this pr small I will handle it in subsequent PRs,

boltdb.go
```diff
@@ -303,27 +310,34 @@ func (itr *boltDBIterator) Valid() bool {
 	return true
 }
 
-func (itr *boltDBIterator) Next() {
-	itr.assertIsValid()
+func (itr *boltDBIterator) Next() error {
+	if err := itr.assertIsValid(); err != nil {
```
melekes: assert should panic I believe

c_level_db.go
```diff
@@ -286,40 +303,55 @@ func (itr cLevelDBIterator) Valid() bool {
 	return true
 }
 
-func (itr cLevelDBIterator) Key() []byte {
-	itr.assertNoError()
-	itr.assertIsValid()
-	return itr.source.Key()
+func (itr cLevelDBIterator) Key() ([]byte, error) {
+	if err := itr.assertNoError(); err != nil {
```
melekes: same. assertX should panic

prefix_db.go
```diff
@@ -267,7 +301,11 @@ type prefixIterator struct {
 }
 
 func newPrefixIterator(prefix, start, end []byte, source Iterator) *prefixIterator {
-	if !source.Valid() || !bytes.HasPrefix(source.Key(), prefix) {
+	// Ignoring the error here as the iterator is invalid
+	// but this is being conveyed in the below if statement
+	key, _ := source.Key() //nolint:errcheck
+
+	if !source.Valid() || !bytes.HasPrefix(key, prefix) {
```
melekes: the better way may be 

```go
pitr := &prefixIterator{
	prefix: prefix,
	start:  start,
	end:    end,
	source: source,
	valid:  false,
}

if !source.Valid() {
	return pitr
}
key, err := source.Key()
if err != nil {
	return nil, err
}
if !bytes.HasPrefix(key, prefix) {
	return pitr
}
```

prefix_db.go
```diff
@@ -329,7 +383,7 @@ func stripPrefix(key []byte, prefix []byte) (stripped []byte) {
 		panic("should not happen")
 	}
 	if !bytes.Equal(key[:len(prefix)], prefix) {
-		panic("should not happne")
+		panic("should not happn")
```
melekes: happen

prefix_db.go
```diff
@@ -293,29 +331,45 @@ func (itr *prefixIterator) Valid() bool {
...
-		return
+		return nil
 	}
+	return nil
 }
 
-func (itr *prefixIterator) Key() (key []byte) {
+func (itr *prefixIterator) Key() (key []byte, err error) {
 	if !itr.valid {
-		panic("prefixIterator invalid, cannot call Key()")
```
melekes: panic is appropriate here as it checks program logic (ie. nobody should call Key/Value if iterator is not Valid)

PR #45, opened 2020-02-01
docs: add readme and test commands

marbar3778:


README.md
```diff
@@ -2,3 +2,53 @@
...
+
+Go 1.13+
+
+## Supported Databases
+
+- [GolevelDB](https://github.com/syndtr/goleveldb)
+- [ClevelDB](https://github.com/google/leveldb)
+- [BoltDB](https://github.com/etcd-io/bbolt)
+- [GoRocksDB](https://github.com/tecbot/gorocksdb)
+- [MemDB](./mem_db.go)
```
melekes:
```suggestion
- [MemDB](#memdb)
```

README.md
```diff
@@ -2,3 +2,53 @@
...
+
+To use goleveldb there is no need to install anything and you can run the tests by simply running `make test`
+
+### ClevelDB
+
+To use cleveldb leveldb must be installed on your machine. Please use you local package manager (brew, snap) to install the db. Once you have installed it you can run tests with `make test-cleveldb`.
+
+### BoltDB
+
+To use BoltDB, you must have it installed on your machine. You can install it by running:
```
melekes: Hmm, is this true? Yes, you can install the binary to manually do compaction, view something, but I thought it's optional...

marbar3778: just tested it, you are right.

melekes: go runtime won't include code in the end binary (`tendermint`)

PR #53, opened 2020-03-03
memdb: use btree for storage

### LINKED ISSUE #52: memdb: use ordered data structure
MemDB currently uses a Go map for storage, which means all ordered operations (such as range scans) are extremely slow - see e.g. https://github.com/tendermint/tendermint/issues/4520.

By using an ordered data structure instead, e.g. a B-tree, these operations would be significantly faster.

erikgrinaker:
Fixes #52 (see also https://github.com/tendermint/tendermint/issues/4520). Uses a B-tree for storage, which should significantly improve range scan performance. No breaking changes.

Haven't had time to benchmark this yet, will do so shortly.

### DIFF
--- a/backend_test.go
+++ b/backend_test.go
@@ -204,6 +204,7 @@ func TestDBIterator(t *testing.T) {
 	for dbType := range backends {
 		t.Run(fmt.Sprintf("%v", dbType), func(t *testing.T) {
 			testDBIterator(t, dbType)
+			testDBIterator_blankKey(t, dbType)
 		})
 	}
 }
@@ -311,6 +312,18 @@ func testDBIterator(t *testing.T, backend BackendType) {
 	verifyIterator(t, ritr,
 		[]int64(nil), "reverse iterator from 7 (ex) to 6")
 
+	ritr, err = db.ReverseIterator(int642Bytes(10), nil)
+	require.NoError(t, err)
+	verifyIterator(t, ritr, []int64(nil), "reverse iterator to 10")
+
+	ritr, err = db.ReverseIterator(int642Bytes(6), nil)
+	require.NoError(t, err)
+	verifyIterator(t, ritr, []int64{9, 8, 7}, "reverse iterator to 6")
+
+	ritr, err = db.ReverseIterator(int642Bytes(5), nil)
+	require.NoError(t, err)
+	verifyIterator(t, ritr, []int64{9, 8, 7, 5}, "reverse iterator to 5")
+
 	// verifyIterator(t, db.Iterator(int642Bytes(0), int642Bytes(1)), []int64{0}, "forward iterator from 0 to 1")
 
 	ritr, err = db.ReverseIterator(int642Bytes(8), int642Bytes(9))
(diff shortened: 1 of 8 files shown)


backend_test.go
```diff
@@ -329,7 +342,56 @@ func testDBIterator(t *testing.T, backend BackendType) {
 	require.NoError(t, err)
 	verifyIterator(t, ritr,
 		[]int64(nil), "reverse iterator from 2 (ex) to 4")
+}
+
+func testDBIterator_blankKey(t *testing.T, backend BackendType) {
```
golangcibot: don't use underscores in Go names; func testDBIterator_blankKey should be testDBIteratorBlankKey (from `golint`)

mem_db.go
```diff
@@ -157,113 +183,145 @@ func (db *MemDB) NewBatch() Batch {
...
+	go func() {
+		// Because we use [start, end) for reverse ranges, while btree uses (start, end], we need
+		// the following variables to handle some reverse iteration conditions ourselves.
+		var (
+			skipEqual     []byte
+			abortLessThan []byte
+		)
+		visitor := func(i btree.Item) bool {
+			item := i.(*item)
+			if skipEqual != nil && bytes.Compare(item.key, skipEqual) == 0 {
```
golangcibot: S1004: should use bytes.Equal(item.key, skipEqual) instead (from `gosimple`)

erikgrinaker:
Benchmarks, doing range scans of 10.000 keys on databases with 1 million and 10 million keys:

| Benchmark                           | Map (ns/op)   | B-tree (ns/op) | Change  |
|-------------------------------------|--------------:|---------------:|--------:|
| `BenchmarkMemDBRangeScans1M-4`      |    90 239 121 |      1 477 749 | -98.36%  |
| `BenchmarkMemDBRangeScans10M-4`     | 1 006 363 039 |      1 469 495 | -99.86% |
| `BenchmarkMemDBRandomReadsWrites-4` |           816 |          3 314 | +306.17%   |

For these database sizes, B-trees have a speedup of 2-3 orders of magnitude. More importantly, and as expected, the old map-based database has a dependence on the number of keys in the database while B-trees do not.

The tradeoff is that read/write operations are three times more costly, but I think that's well worth it.

common_test.go
```diff
@@ -217,23 +249,27 @@ func benchmarkRandomReadsWrites(b *testing.B, db DB) {
...
-			idx := int64(rand.Int()) % numItems // nolint:gosec testing file, so accepting weak random number generator
+			idx := rand.Int63n(numItems) // nolint:gosec testing file, so accepting weak random number generator
 			internal[idx]++
 			val := internal[idx]
 			idxBytes := int642Bytes(idx)
 			valBytes := int642Bytes(val)
 			//fmt.Printf("Set %X -> %X\n", idxBytes, valBytes)
 			err := db.Set(idxBytes, valBytes)
-			b.Error(err)
+			if err != nil {
```
melekes: not sure why you need `if`

common_test.go
```diff
@@ -217,23 +249,27 @@ func benchmarkRandomReadsWrites(b *testing.B, db DB) {
...
 
 		// Read something
 		{
-			idx := int64(rand.Int()) % numItems // nolint:gosec testing file, so accepting weak random number generator
+			idx := rand.Int63n(numItems) // nolint:gosec testing file, so accepting weak random number generator
 			valExp := internal[idx]
 			idxBytes := int642Bytes(idx)
 			valBytes, err := db.Get(idxBytes)
-			b.Error(err)
+			if err != nil {
```
melekes: same

mem_db.go
```diff
@@ -15,40 +40,42 @@ func init() {
...
-	mtx sync.Mutex
-	db  map[string][]byte
+	mtx   sync.RWMutex
+	btree *btree.BTree
 }
 
 func NewMemDB() *MemDB {
 	database := &MemDB{
-		db: make(map[string][]byte),
+		btree: btree.New(32),
```
melekes: what 32 is here?

mem_db.go
```diff
@@ -157,113 +183,144 @@ func (db *MemDB) NewBatch() Batch {
...
-func newMemDBIterator(db DB, keys []string, start, end []byte) *memDBIterator {
-	return &memDBIterator{
-		db:    db,
-		cur:   0,
-		keys:  keys,
-		start: start,
-		end:   end,
+func newMemDBIterator(bt *btree.BTree, start []byte, end []byte, reverse bool) *memDBIterator {
+	ctx, cancel := context.WithCancel(context.Background())
+	ch := make(chan *item, 64)
```
melekes: why 64? maximum 64 items?

common_test.go
erikgrinaker: Performance. `require.NoError()` is very costly, so without the `if` the benchmark time is dominated by `require.NoError()` (profiler said about 60%, if I remember correctly).

mem_db.go
erikgrinaker: The "degree" of the B-tree, i.e. the number of items per tree node. Benchmarks showed 32 to be the sweet spot.

mem_db.go
erikgrinaker: We buffer 64 items in the channel. This allows the goroutines to do more work per context switch. If the channel is unbuffered, then you get 2 context switches per item, which kills performance. Benchmarks showed 64 to be the sweet spot.

I'll add constants for these values.

common_test.go
melekes: kk. then why not use if err != nil { b.Fatal(err) } ?

mem_db.go
melekes: can we add a const degree = 32 and a comment saying 32 is best according to benchmarks?

common_test.go
```diff
@@ -213,7 +213,8 @@ func benchmarkRangeScans(b *testing.B, db DB, dbSize int64) {
 		bytes := int642Bytes(i)
 		err := db.Set(bytes, bytes)
 		if err != nil {
-			// for performance
+			// require.NoError() is very expensive (according to profiler), so we do a cheap if as
+			// well since this is a tight loop.
 			require.NoError(b, err)
```
melekes: b.Fatal(err)

common_test.go
erikgrinaker: Because require.NoError() generates nice output automatically. :) But sure, I'll update.

mem_db.go
erikgrinaker: Yep, already done

erikgrinaker:
I ran some Cosmos SDK benchmarks using the B-tree MemDB, via `make test-sim-benchmark`. These used `SIM_NUM_BLOCKS=250`, `SIM_BLOCK_SIZE=200`, and LevelDB with pruning strategies `PruneNothing` (persist every version to disk) and `PruneSyncable` (persist every 100 blocks to disk). I believe the SDK currently uses `PruneSyncable` by default.

| `tm-db` | `PruneNothing` | `PruneSyncable` | `MemDB` + `PruneNothing` |
|---------|----------------|-----------------|--------------------------|
| 0.4.1   | 191 s          | 114 s           | 691 s                    |
| btree   | 189 s          | 117 s           | 138 s                    |

It appears that IAVLs use of MemDB in `PruneSyncable` mode, as a temporary working set that's periodically flushed to disk, does not trigger the slow path in MemDB. It is unclear why, but appears to be related to cleanup of orphaned nodes, which possibly hits LevelDB instead of MemDB. In this case the new B-tree appears to have a slight slowdown, as expected. However, when we use MemDB as the main database we do hit the slow path, and the B-tree version is five times faster.

Overall memory usage is about the same in both cases.

PR #56, opened 2020-03-04
memdb: API cleanups and improvements

### LINKED ISSUE #54: Reduce MemDB API surface
MemDB currently has lots of exported methods that are clearly meant for internal use, e.g. `Mutex()`, `SetNoLock()`, and `DeleteNoLock()`. These are primarily used by `memBatch`. We should make these internal.

### LINKED ISSUE #55: Use RWMutex in MemDB
MemDB currently uses a `Mutex`, but it should use an `RWMutex` instead. However, the mutex is returned through the public `Mutex()` method, so it's a breaking change. We should unexport `Mutex()` in #54 and then change it.

erikgrinaker:
Fixes #54, fixes #55. Somewhat breaking, since we're removing exported methods, but these are not likely to be widely used.

* Removes atomicSetDeleter interface and related public methods
* Uses an `RWMutex` instead of a `Mutex` in MemDB
* Removes `mockDB` since it is not really used
* Adds batch tests for all backends
* Various minor code cleanups

Branched off of #53, should be rebased onto `master` before merging.


mem_db.go
```diff
@@ -335,3 +302,61 @@ func (i *memDBIterator) Value() []byte {
 	}
 	return i.item.value
 }
+
+// memBatch operations
+type opType int
+
+const (
+	opTypeSet opType = iota
```
melekes: https://github.com/uber-go/guide/blob/master/style.md#start-enums-at-one

mem_db.go
```diff
@@ -335,3 +302,61 @@ func (i *memDBIterator) Value() []byte {
...
+	defer mBatch.db.mtx.Unlock()
+
+	for _, op := range mBatch.ops {
+		switch op.opType {
+		case opTypeSet:
+			mBatch.db.set(op.key, op.value)
+		case opTypeDelete:
+			mBatch.db.delete(op.key)
+		default:
+			return errors.Errorf("unknown operation %T", op)
```
melekes:
```suggestion
			return errors.Errorf("unknown operation type %T (%v)", opType, op)
```

mem_db.go
erikgrinaker: Good call, thanks

mem_db.go
```diff
@@ -189,25 +158,24 @@ func (db *MemDB) NewBatch() Batch {
...
-	db.mtx.Lock()
-	defer db.mtx.Unlock()
-
+	db.mtx.RLock()
+	defer db.mtx.RUnlock()
 	return newMemDBIterator(db.btree, start, end, true), nil
 }
 
+//----------------------------------------
+// Iterator
```
tessr: We've been talking for a long time about getting rid of these ascii separators and instead putting things in their own files. Do you want to do that while you're in here cleaning things up? (Could also go in a follow up PR, since it might create a lot of noise in this one.)

tessr:
Nice. Happy to see an unused DB removed!

mem_db.go
erikgrinaker: Sure thing

erikgrinaker: I split the iterator and batch into separate files, but kept them in the root directory since using a separate directory would break the API quite severely. I also renamed files from e.g. `mem_db_iterator.go` to `memdb_iterator.go`.

Let me know if you think this is fine, and I'll merge this and submit a PR that does the same for the other backends.

memdb_batch.go
```diff
@@ -0,0 +1,63 @@
...
+	defer b.db.mtx.Unlock()
+
+	for _, op := range b.ops {
+		switch op.opType {
+		case opTypeSet:
+			b.db.set(op.key, op.value)
+		case opTypeDelete:
+			b.db.delete(op.key)
+		default:
+			return errors.Errorf("unknown operation type %T (%v)", op.opType, op)
```
melekes:
```suggestion
			return errors.Errorf("unknown operation type %v (%v)", op.opType, op)
```

PR #58, opened 2020-03-04
Make GoLevelDB batch.Close() actually close

### LINKED ISSUE #57: GoLevelDB's batch.Close() does not actually close
See failing test here: https://github.com/tendermint/tm-db/pull/56/files#diff-036725878b6bdf47f6fa6935794c1cceR456

erikgrinaker:
Fixes #57. Branched off of #56, rebase onto `master` once merged.


tessr:
Nice. (Why was this a no-op previously?? 🤔)

erikgrinaker:
> Nice. (Why was this a no-op previously?? 🤔)

Probably just an oversight. The only case where this matters is when the batch is reused, and I don't think that's very common.

PR #64, opened 2020-03-09
circleci: run all tests

erikgrinaker:
Test all backends in CircleCI. Currently fails, awaiting fixes in #65, #66, and #69.


marbar3778:
Im guessing this is failing correctly now 😄

PR #65, opened 2020-03-09
cleveldb: fix handling of empty iterator keys

erikgrinaker:
New tests showed that empty iterator keys causes panics.


erikgrinaker:
@melekes Changed this to only fix the iterator bug, while I make a new PR for closed batch errors. Could you review again please?

PR #66, opened 2020-03-09
rocksdb: handle reuse of closed batches

erikgrinaker:
Fixes a bug uncovered by new tests.

### DIFF
--- a/rocksdb_batch.go
+++ b/rocksdb_batch.go
@@ -11,35 +11,51 @@ type rocksDBBatch struct {
 
 var _ Batch = (*rocksDBBatch)(nil)
 
+func newRocksDBBatch(db *RocksDB) *rocksDBBatch {
+	return &rocksDBBatch{
+		db:    db,
+		batch: nil,
+	}
+}
+
+func (b *rocksDBBatch) ensureBatch() {
+	if b.batch == nil {
+		b.batch = gorocksdb.NewWriteBatch()
+	}
+}
+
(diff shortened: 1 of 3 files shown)


rocksdb_batch.go
```diff
@@ -11,35 +11,51 @@ type rocksDBBatch struct {
...
 }
 
 // Write implements Batch.
-func (mBatch *rocksDBBatch) Write() error {
-	err := mBatch.db.db.Write(mBatch.db.wo, mBatch.batch)
-	if err != nil {
-		return err
+func (b *rocksDBBatch) Write() error {
+	if b.batch == nil {
+		return nil
```
melekes: I believe we need to return an error instead of silently ignoring write. we should help developer uncover his/her errors (not executing Set/Delete on batch), not ignoring them.

erikgrinaker: Then why did you approve #67? :) This can only be `nil` if we write a closed or empty batch, in which case we need to silently ignore the write.

melekes: > This can only be nil if we write a closed or empty batch, in which case we need to silently ignore the write to satisfy the semantics tested by #67. 

can't you modify #67 if you agree with my logic?

erikgrinaker: Sure, I'll submit a new PR. It will be a bit more breaking, because we have to change the `Batch` interface to return errors etc, but I agree it's a better solution.

PR #68, opened 2020-03-09
test: properly close iterator in assertKeyValues()

erikgrinaker:
Fixes deadlocks with backends that have locking iterators (namely BoltDB).


backend_test.go
```diff
@@ -504,6 +504,7 @@ func assertKeyValues(t *testing.T, db DB, expect map[string][]byte) {
 		require.NoError(t, iter.Error())
 		actual[string(iter.Key())] = iter.Value()
 	}
+	iter.Close()
```
melekes: better to move it to 501 (+ defer) because 504 can return early?

erikgrinaker: Sure

PR #69, opened 2020-03-09
boltdb: handle empty key placeholders in iterators

erikgrinaker:
Iterators did not handle the placeholder `"nil"` key used for empty keys (related to #2)

### DIFF
--- a/boltdb_iterator.go
+++ b/boltdb_iterator.go
@@ -13,9 +13,10 @@ import (
 type boltDBIterator struct {
 	tx *bbolt.Tx
 
-	itr   *bbolt.Cursor
-	start []byte
-	end   []byte
+	itr           *bbolt.Cursor
+	start         []byte
+	end           []byte
+	emptyKeyValue []byte // Tracks the value of the empty key, if it exists
 
 	currentKey   []byte
 	currentValue []byte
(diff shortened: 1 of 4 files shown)


boltdb_iterator.go
```diff
@@ -28,33 +29,58 @@ var _ Iterator = (*boltDBIterator)(nil)
...
 			_, _ = itr.Seek(end) // after key
 			ck, cv = itr.Prev()  // return to end key
 		}
+		// If we're currently positioned at the placeholder for the empty key, skip it (handle later)
+		if emptyKeyValue != nil && bytes.Equal(ck, boltDBEmptyKey) {
+			ck, cv = itr.Prev()
+		}
+		// If we didn't find any initial key, but there's a placeholder for the empty key at the
+		// end that we've skipped, then the initial key should be the empty one (the final one).
+		if emptyKeyValue != nil && ck == nil && (end == nil || len(end) > 0) {
```
melekes: `(end == nil || len(end) > 0)` can't we rewrite this as `len(end) == 0`?

erikgrinaker: No, what we specifically want to check is that `end` is not `[]byte{}`. `len(end) != 0` does not work, because `len(nil) == 0`.

PR #71, opened 2020-03-10
Calling non-Close() methods on closed or written batches panic

erikgrinaker:
See discussion in https://github.com/tendermint/tm-db/pull/66#pullrequestreview-371460117.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 - Closed batches can no longer be reused, all non-`Close()` calls return `ErrBatchClosed`
 
+- Calling `Write()` or `WriteSync()` on a `Batch` will implicitly close it
+
 - Removed the `SetDeleter` interface
 
 - [memdb] [\#56](https://github.com/tendermint/tm-db/pull/56) Removed some exported methods that were mainly meant for internal use: `Mutex()`, `SetNoLock()`, `SetNoLockSync()`, `DeleteNoLock()`, and `DeleteNoLockSync()`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,7 @@
 
 ### Breaking Changes
 
-- `Batch` interface methods now return errors
-
-- Closed batches can no longer be reused, all non-`Close()` calls return `ErrBatchClosed`
-
-- Calling `Write()` or `WriteSync()` on a `Batch` will implicitly close it
-
-- Removed the `SetDeleter` interface
+- [\#71](https://github.com/tendermint/tm-db/pull/71) Closed or written batches can no longer be reused, all non-`Close()` calls will panic
 
 - [memdb] [\#56](https://github.com/tendermint/tm-db/pull/56) Removed some exported methods that were mainly meant for internal use: `Mutex()`, `SetNoLock()`, `SetNoLockSync()`, `DeleteNoLock()`, and `DeleteNoLockSync()`
 
@@ -22,9 +16,9 @@
 
 ### Bug Fixes
 
-- [boltdb] Properly handle blank keys in iterators
+- [boltdb] [\#69](https://github.com/tendermint/tm-db/pull/69) Properly handle blank keys in iterators
 
-- [cleveldb] Fix handling of empty keys as iterator endpoints
+- [cleveldb] [\#65](https://github.com/tendermint/tm-db/pull/65) Fix handling of empty keys as iterator endpoints
 
 ## 0.4.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - `Batch` interface methods now return errors
 
-- Closed batches can no longer be reused
+- Closed batches can no longer be reused, all non-`Close()` calls return `ErrBatchClosed`
 
 - Removed the `SetDeleter` interface
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - `Batch` interface methods now return errors
 
+- Closed batches can no longer be reused
+
 - Removed the `SetDeleter` interface
 
 - [memdb] [\#56](https://github.com/tendermint/tm-db/pull/56) Removed some exported methods that were mainly meant for internal use: `Mutex()`, `SetNoLock()`, `SetNoLockSync()`, `DeleteNoLock()`, and `DeleteNoLockSync()`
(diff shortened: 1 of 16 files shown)


boltdb_batch.go
```diff
@@ -12,19 +12,37 @@ type boltDBBatch struct {
...
+		db:  db,
+		ops: make([]operation, 0, 1),
+	}
+}
+
 // Set implements Batch.
-func (b *boltDBBatch) Set(key, value []byte) {
+func (b *boltDBBatch) Set(key, value []byte) error {
+	if b.ops == nil {
+		return ErrBatchClosed
```
melekes: what do u think about panicking here?

cleveldb_batch.go
```diff
@@ -10,33 +10,60 @@ type cLevelDBBatch struct {
...
-	if err := b.db.db.Write(b.db.wo, b.batch); err != nil {
+	if b.batch == nil {
+		return ErrBatchClosed
+	}
+	err := b.db.db.Write(b.db.wo, b.batch)
+	if err != nil {
 		return err
 	}
-	return nil
+	return b.Close()
```
melekes: do we even need a separate Close method? if we execute Close in Write/WriteSync

erikgrinaker: Yeah, because the caller may e.g. return an error while they are building the batch, in which case they need to close it without calling `Write`.

boltdb_batch.go
erikgrinaker: Yeah, I was thinking about that too. Not really sure - I don't like all of the panicing we do, but in this case it might be appropriate: we wouldn't need to change the interface, and callers wouldn't have to check for errors all the time.

The standard library does both: writing to a closed channel panics, but writing to a closed `io.Writer` returns an error.

@tessr @marbar3778 What do you guys think?

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@
 
 ### Breaking Changes
 
+- `Batch` interface methods now return errors
```
marbar3778: missing pr numbers on changelog entries

cleveldb_batch.go
melekes: I can easily envision cases where Close will be called twice... maybe it's okay, but I'd prefer to remove Close calls from Write/WriteSync. You're mixing responsibilities

erikgrinaker: Calling `Close()` twice is fine, it is idempotent. It is a normal Go pattern to use `defer c.Close()` and then also do an explicit `c.Close()` as early as possible to free up resources before the function exits.

If we don't call `Close()` in `Write()`, then what should happen when a caller calls `Write()` twice?

boltdb_batch.go
erikgrinaker: Discussed this with @marbar3778 on Slack. Breaking the interface here doesn't actually buy us anything useful, since none of the Tendermint/SDK/IAVL uses of `Batch` are affected by this. I suggest we panic for now, and consider changing the interface when we have to break it for an actually useful reason.

cleveldb_batch.go
melekes: > It is a normal Go pattern to use defer c.Close() for error handling and then also do an explicit c.Close() as early as possible to free up resources before the function exits.

Didn't know about such.

> If we don't call Close() in Write(), then what should happen when a caller calls Write() twice?

you can't prevent people from shooting themselves into the leg. I'd assume correct usage in most cases.

New -> Set/Delete -> Write(Sync) -> Close (via defer)

erikgrinaker: > you can't prevent people from shooting themselves into the leg. I'd assume correct usage in most cases.

Of course, we just have to specify what should happen, as unspecified behavior is the root of all evil. Either `Write()` also calls `Close()`, in which case a second `Write()` call errors, otherwise the second `Write()`call should write the same batch again.

I think erroring is safer, and easier to implement across all backends. The typical and recommended usage would still be the same flow that you outline.

melekes: > I think erroring is safer, and easier to implement across all backends. 

Agree.

CHANGELOG.md
erikgrinaker: Thanks, fixed

backend_test.go
```diff
@@ -439,41 +439,14 @@ func testDBBatch(t *testing.T, backend BackendType) {
...
-	batch.Close()
-
-	batch.Delete([]byte("c"))
-	err = batch.WriteSync()
-	require.NoError(t, err)
-	assertKeyValues(t, db, map[string][]byte{"a": {1}, "b": {2}})
+	// trying to modify or rewrite a written batch should panic, but closing it should work
+	require.Panics(t, func() { batch.Set([]byte("a"), []byte{9}) })
+	require.Panics(t, func() { batch.Delete([]byte("a")) })
+	require.Panics(t, func() { batch.Write() })
```
golangcibot: Error return value of `batch.Write` is not checked (from `errcheck`)

backend_test.go
```diff
@@ -439,41 +439,14 @@ func testDBBatch(t *testing.T, backend BackendType) {
...
+	require.Panics(t, func() { batch.WriteSync() })
```
golangcibot: Error return value of `batch.WriteSync` is not checked (from `errcheck`)

backend_test.go
```diff
@@ -486,13 +459,38 @@ func testDBBatch(t *testing.T, backend BackendType) {
...
+
+	// it should be possible to close an empty batch, and to re-close a closed batch
+	batch = db.NewBatch()
+	batch.Close()
+	batch.Close()
+
+	// all other operations on a closed batch should panic
+	require.Panics(t, func() { batch.Set([]byte("a"), []byte{9}) })
+	require.Panics(t, func() { batch.Delete([]byte("a")) })
+	require.Panics(t, func() { batch.Write() })
```
golangcibot: Error return value of `batch.Write` is not checked (from `errcheck`)

backend_test.go
```diff
@@ -486,13 +459,38 @@ func testDBBatch(t *testing.T, backend BackendType) {
...
+	require.Panics(t, func() { batch.WriteSync() })
```
golangcibot: Error return value of `batch.WriteSync` is not checked (from `errcheck`)

cleveldb_batch.go
erikgrinaker: Ok, I think this should be good to go then.

boltdb_batch.go
```diff
@@ -12,19 +12,35 @@ type boltDBBatch struct {
 
 var _ Batch = (*boltDBBatch)(nil)
 
+func newBoltDBBatch(db *BoltDB) *boltDBBatch {
+	return &boltDBBatch{
+		db:  db,
+		ops: make([]operation, 0, 1),
```
melekes: how does setting cap=1 helps?

erikgrinaker: It basically doesn't. The reasoning was that users would always create a batch to put at least 1 item in it, so 1 is more appropriate than 0, but it's not going to have any real-world implications so I've changed it to an empty slice.

PR #77, opened 2020-03-14
switch from golangci to github actions

marbar3778:
- golangci service is being deprecated:
https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e

transition to githubactions run linting


erikgrinaker:
Oh no. ☹️ It might be easier to just run this in CircleCI, to avoid having to maintain too many different build systems.

marbar3778:
was thinking of doing that as well, but think this may be simpler than waiting to add a token to our circle instance. I would also like to look at moving this repo entirely over to github actions

PR #85, opened 2020-04-02
use Apache 2.0 license

### LINKED ISSUE #84: Lack of a license
It seems like there is no LICENSE file in this repository. Assuming this was meant to be open source, could you please add one?

erikgrinaker:
Fixed #84. I'm assuming this is appropriate. Snagged from Tendermint repo.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,204 @@
+Tendermint Core
+License: Apache2.0
+
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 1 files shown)


alessio:
This repository was created out of the old github.com/tendermint/tendermint repository.
The code has always been distributed under the terms of the Apache-2.0 license, thus to the best of my knowledge (and with my Debian Developer hat on) this action is legit and does not even qualify as sublicensing. Hence merging.

PR #96, opened 2020-05-15
batch: return errors for all operations

erikgrinaker:
Modify batch `Set`, `Delete`, and `Close` to return errors. The main error condition currently is using closed batches, but future ones include databases which implement batches as transactions (e.g. BadgerDB) and rejecting empty keys (#2).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Breaking Changes
+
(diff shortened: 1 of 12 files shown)


marbar3778:
Lint is failing due to errors in tests, we can turn off linting in tests if we would like

erikgrinaker:
> Lint is failing due to errors in tests, we can turn off linting in tests if we would like

That's fine, I just couldn't find the actual lint error. Will fix shortly.

PR #99, opened 2020-05-18
reject nil/empty keys and nil values

### LINKED ISSUE #2: libs/db: disallow empty keys
See https://github.com/tendermint/tendermint/pull/3610#discussion_r280053143

Reasons to do that:
- many KV stores does not support empty keys
- I don't know why would someone want to set en empty key (and not call it `default` or something else)

erikgrinaker:
Fixes #2.


melekes:
what's the reason behind rejecting empty values? I wasn't able to find anything in the linked issue nor here.

PR #100, opened 2020-05-28
feat: add readonly mode for goleveldb

cpucorecore:
the readonly mode can speedu getting keys from keybase without use lock.

### DIFF
--- a/db.go
+++ b/db.go
@@ -14,6 +14,8 @@ const (
 	//   - pure go
 	//   - stable
 	GoLevelDBBackend BackendType = "goleveldb"
+	// same to GoLevelDBBackend but readonly
+	GoLevelDBBackendReadOnly BackendType = "goleveldb_readonly"
 	// CLevelDBBackend represents cleveldb (uses levigo wrapper)
 	//   - fast
 	//   - requires gcc
--- a/goleveldb.go
+++ b/goleveldb.go
@@ -14,6 +14,11 @@ func init() {
 		return NewGoLevelDB(name, dir)
 	}
 	registerDBCreator(GoLevelDBBackend, dbCreator, false)
+
+	dbCreatorReadOnly := func(name string, dir string) (DB, error) {
+		return NewGoLevelDBReadOnly(name, dir)
+	}
+	registerDBCreator(GoLevelDBBackendReadOnly, dbCreatorReadOnly, false)
 }
 
 type GoLevelDB struct {
@@ -24,6 +29,12 @@ var _ DB = (*GoLevelDB)(nil)
 
 func NewGoLevelDB(name string, dir string) (*GoLevelDB, error) {
 	return NewGoLevelDBWithOpts(name, dir, nil)
+}
+
+func NewGoLevelDBReadOnly(name string, dir string) (*GoLevelDB, error) {
+	var o opt.Options
+	o.ReadOnly = true
+	return NewGoLevelDBWithOpts(name, dir, &o)
 }
 
 func NewGoLevelDBWithOpts(name string, dir string, o *opt.Options) (*GoLevelDB, error) {


goleveldb.go
```diff
@@ -26,6 +31,12 @@ func NewGoLevelDB(name string, dir string) (*GoLevelDB, error) {
 	return NewGoLevelDBWithOpts(name, dir, nil)
 }
 
+func NewGoLevelDBReadOnly(name string, dir string) (*GoLevelDB, error) {
+	var o opt.Options
+	o.ReadOnly = true
+	return NewGoLevelDBWithOpts(name, dir, &o)
+}
+
```
marbar3778, line 39: thanks for making a contribution. Could you give some context on why this should be added over passing the option to `NewGoLevelDBWithOpts(..)`?

db.go
```diff
@@ -14,6 +14,8 @@ const (
 	//   - pure go
 	//   - stable
 	GoLevelDBBackend BackendType = "goleveldb"
+	// same to GoLevelDBBackend but readonly
+	GoLevelDBBackendReadOnly BackendType = "goleveldb_readonly"
```
tessr, line 18: Can you also explain why you need this change? Thank you!

goleveldb.go
cpucorecore: thank u very much for reply. the context:

Our blockchain project (https://github.com/netcloth/netcloth-chain) client needs to be able to sign transactions with high concurrency. There are thousands of local accounts stored in the tm-db with store type goleveldb, but tm-db does not provide APIs that can set goleveldb read-only mode nor does it support setting goleveldb parameters in db instance constructor. The read TPS is about 20 in non-read-only mode by tm-db.

@tessr

cpucorecore: and The database will not be modified when signing transactions

db.go
cpucorecore: thank u for reply.
tm-db creates the dbCreator list when the module is init, this can easily select the readonly dbCreator

cpucorecore: maybe i should make a issue

goleveldb.go
marbar3778: ah I see its because when we call newDB we are essentially prohibiting any DB options to be added. 

what about adding an options arg into here:
`func NewDB(name string, backend BackendType, dir string, opts interface{}) (DB, error) {`

then in each db we check this opts interface to the appropriate DB's options type if its the wrong type we panic or error. this way users cold pass in options. Thoughts? 

cc @erikgrinaker

cpucorecore: ok， thank u

marbar3778: oh sorry before you close this I would like to get eriks thoughts as he has been doing most of the work in this repo.

cpucorecore: yep, This means that the API interface has been modified, and for the client, all the places that call NewDB need to be modified. A new interface can be added for future clients to use. For the old client code, maybe this PR can help

cpucorecore: client code:

func NewLevelDB(name, dir string) (db dbm.DB, err error) {
	backend := dbm.GoLevelDBBackend
	if DBBackend == string(dbm.CLevelDBBackend) {
		backend = dbm.CLevelDBBackend
	}

	if strings.Contains(dir, ".nchcli") { // nchcli can readonly for signing txs
		backend = dbm.GoLevelDBBackendReadOnly
	}

	defer func() {
		if r := recover(); r != nil {
			err = fmt.Errorf("couldn't create db: %v", r)
		}
	}()
	return dbm.NewDB(name, backend, dir), err
}

// DBBackend is a module variable

erikgrinaker: We already have backend-specific constructors that allow passing options to the backend database, including read-only support for GoLevelDB:

```go
import (
	tmdb "github.com/tendermint/tm-db"
	"github.com/syndtr/goleveldb/leveldb/opt"
)

db, err := tmdb.NewGoLevelDBWithOpts("gldb", "/tmp", opt.Options{ReadOnly: true})
```

Is this not sufficient?

cpucorecore: ok, thank u.

PR #103, opened 2020-06-02
add codecov tags 

marbar3778:


README.md
```diff
@@ -1,8 +1,31 @@
-# Tendermint DB
 
-![Test](https://github.com/tendermint/tm-db/workflows/Test/badge.svg?branch=master)
-![Lint](https://github.com/tendermint/tm-db/workflows/Lint/badge.svg?branch=master)
-[![](https://godoc.org/github.com/tendermint/tm-db?status.svg)](http://godoc.org/github.com/tendermint/tm-db)
+<div align="center">
```
tessr, line 2: I actually think the readme would be easier to read if it weren't centered (and also easier to edit if it were standard markdown rather than html). But I like the design-y intention!

PR #106, opened 2020-06-08
goleveldb: reduce useage

marbar3778:
this is a test pr based off some reading into goleveldb, it still needs to have some testing done. 

Should this be here on the application?


erikgrinaker:
>Should this be here on the application?

I'd say this should be managed by the application - these changes can have pretty significant effects on performance characteristics, and we probably don't want to use non-standard behavior by default.

marbar3778:
then will do this in tendermint when testing

Issue #38, opened 2020-01-09
Supported database backends


marbar3778:
Would like to have a discussion on which dbs we should support in this repo. 

- which DB's are used?

- Which should we deprecate?

melekes:
rocksdb (fast kv store), boltdb (b+tree - useful for indexing), memdb (useful for testing) are my top picks.

we should probably add badger

not a fan of leveldb

> Which should we not maintain anymore?

fsdb?

tessr:
> we should probably add badger

what motivates this? I'd generally like to see requests from users (or prospective users) before adding another database

melekes:
> I'd generally like to see requests from users (or prospective users) before adding another database

I somewhat disagree. From validators (average users) point of view, they usually want stable performant DB with adequate storage requirements. Same for developers building on top of Tendermint, with the exception that they have time / desire sometimes to investigate which DB is the best.

> we should probably add badger
> what motivates this? 

https://github.com/tendermint/tendermint/issues/1835#issuecomment-421062709
either badger or bbolt should be the default, not goleveldb in my opinion.

marbar3778:
>I somewhat disagree. From validators (average users) point of view, they usually want stable performant DB with adequate storage requirements. Same for developers building on top of Tendermint, with the exception that they have time / desire sometimes to investigate which DB is the best. In my mind it's not the matter of adding or removing, but providing best possible option for our users.

I do agree with this, and I am actually in favor of adding badgerdb, but would like to go through some testing. 

Also, another concern of mine that I found in working on the error handling is that the tests are not run on all the dbs, I will spend time seeing what is leading to this, but would like to see what I can get rid of first so I don't spend time on them.

tessr:
> In my mind it's not the matter of adding or removing, but providing best possible option for our users. 

I can agree with this--but I'd like us to make sure we have a manageable maintenance load!

mvdan:
@melekes are you still interested in adding support for badger/v2? Is https://gist.github.com/melekes/85b8c07dd917828dc2ac6696129b73f2 your latest attempt, or has someone else continued the work since?

I'd like to use badger with the iavl library, so I could work on a pull request here if it would be welcome.

marbar3778:
[...]

Issue #62, opened 2020-03-09
Specify and test concurrency and transaction semantics


erikgrinaker:
We should specify and test the concurrency and transaction semantics both of individual interface methods and specific backends. For backends that support it, we should aim to support single-writer-multiple-reader concurrency control (e.g. `RWMutex` unless backend handles this itself), and atomic batch transactions with read committed isolation. As far as possible, iterators should be concurrency safe and not block writers for their entire lifetime.

We may want to consider exposing backend capabilities for this, such that users can programmatically detect e.g. whether a batch write is atomic or not and take appropriate action.

Some related notes: https://github.com/tendermint/tmlibs/issues/79

PR #115, opened 2020-06-30
add a BadgerDB backend

mvdan:
With multiple commits, to show the incremental changes on top of Anton's work in
early 2019.

Updates #38.


mvdan:
FYI @melekes @erikgrinaker @marbar3778

mvdan:
CI is failing because interfacer is complaining about an unexported function. You should disable that linter altogether - that warning isn't particularly useful :)

badger_db.go
```diff
@@ -0,0 +1,297 @@
...
+)
+
+func init() { registerDBCreator(BadgerDBBackend, badgerDBCreator, true) }
+
+func badgerDBCreator(dbName, dir string) (DB, error) {
+	return NewBadgerDB(dbName, dir)
+}
+
+// NewBadgerDB creates a Badger key-value store backed to the
+// directory dir supplied. If dir does not exist, we create it.
```
melekes, line 19:
```suggestion
// directory dir supplied. If dir does not exist, it will be created.
```

badger_db.go
```diff
@@ -0,0 +1,297 @@
...
+	if len(key) == 0 {
+		return errKeyEmpty
+	}
+	return b.db.Update(func(txn *badger.Txn) error {
+		return txn.Delete(key)
+	})
+}
+
+func (b *BadgerDB) DeleteSync(key []byte) error {
+	if len(key) == 0 {
```
melekes, line 121:
```suggestion
	return withSync(b.db, b.Delete(key))
```

mvdan:
badger_db.go is under 300 lines, so I would generally not split it up. The current file structure seems a bit weird to me, with over a dozen files under 100 lines, and over 30 files total. But if you strongly prefer consistency, I can split it up.

mvdan:
Two follow-up comments:

1) The tests here are passing, but I assume something will fail once tendermint is properly tested with this code. Can someone with more tendermint experience help?

2) It would be interesting to get some benchmark numbers. I can help if you can point me at a benchmark that is bottlenecked by the database, and that reflects somewhat realistic usage.

marbar3778:
> badger_db.go is under 300 lines, so I would generally not split it up. The current file structure seems a bit weird to me, with over a dozen files under 100 lines, and over 30 files total. But if you strongly prefer consistency, I can split it up.

it should be fine since there will be a followup to restructuring the package.

badger_db.go
```diff
@@ -0,0 +1,292 @@
...
+	return true
+}
+
+func (i *badgerDBIterator) Key() []byte {
+	if !i.Valid() {
+		panic("iterator is invalid")
+	}
+	// Note that we don't use KeyCopy, so this is only valid until the next
+	// call to Next.
+	return i.iter.Item().Key()
```
erikgrinaker, line 280: We should use `KeyCopy()` here. The tm-db API only specifies that the key slice cannot be modified, so we shouldn't introduce backend-specific limitations like these as they can suddenly break applications when the user changes the backend.

badger_db.go
```diff
@@ -0,0 +1,292 @@
...
+
+func (b *BadgerDB) iteratorOpts(start, end []byte, opts badger.IteratorOptions) (*badgerDBIterator, error) {
+	if (start != nil && len(start) == 0) || (end != nil && len(end) == 0) {
+		return nil, errKeyEmpty
+	}
+	txn := b.db.NewTransaction(false)
+	iter := txn.NewIterator(opts)
+	iter.Rewind()
+	iter.Seek(start)
+	if opts.Reverse && bytes.Equal(iter.Item().Key(), start) {
```
erikgrinaker, line 140: I believe we need to check `iter.Valid()` here as well.

badger_db.go
```diff
@@ -0,0 +1,292 @@
...
+
+func (b *BadgerDB) Stats() map[string]string {
+	return nil
+}
+
+func (b *BadgerDB) NewBatch() Batch {
+	wb := &badgerDBBatch{
+		db:         b.db,
+		wb:         b.db.NewWriteBatch(),
+		firstFlush: make(chan struct{}, 1),
```
erikgrinaker, line 174: What is the advantage of using a channel over a simple boolean?

badger_db.go
mvdan: Thanks - I thought about it, but since the interface didn't specify that the `[]byte` are safe to use forever, I leaned towards the simpler (less expensive) option and the comment.

But what you say makes sense. I'll swap it. It's unfortunate that the interface has this performance limitation, though :)

badger_db.go
mvdan: A simple boolean would be racy, assuming that there can be concurrent method calls.

mvdan: To clarify - I could use a bool field and atomic reads/writes, and perhaps that would be a way to avoid the allocation in the batch constructor. If you think that's worth the effort, I can look into it. I assumed that a write batch will generally do a considerable amount of work, so a single allocation shouldn't be noticeable.

badger_db.go
mvdan: ACK, thanks!

badger_db.go
erikgrinaker: I agree, and scans are particularly performance-sensitive. We can consider adding this requirement, but it would require vetting a bunch of existing code. It would be useful with a realistic benchmark to back this change.

mvdan:
> Can you add a README description and changelog entry as well please?

For sure.

> The Tendermint end-to-end tests pass, but IAVL fails with a panic.

Thanks, I'll take a look.

> The best option would probably be Cosmos SDK simulations, which use tm-db via IAVL. I'm not very familiar with these though, as it's managed by a different team.

Sounds good - I'll wait for someone from that team to give some insight, then. Benchmarking should probably come once the reviews are done.

badger_db.go
mvdan: Yeah, it sounds like a change that should be considered carefully at a later time. Definitely not for this PR - just thinking outloud.

badger_db.go
erikgrinaker: Right. tm-db currently doesn't specify any concurrency semantics (#62), and the primary users (e.g. IAVL) are not concurrent. But doesn't hurt to be safe, I suppose. I personally find an atomic expresses the intent better, but it's not a big deal - a channel will do just fine.

erikgrinaker:
@alexanderbez Do you have any SDK sims that would be suitable for benchmarking database backends with realistic workloads?

alexanderbez:
Only thing that comes to mind is simulation benchmarking, which you can look into. Although, you should be able to get away with a pretty easy one-off benchmark, where you create a simapp with a corresponding backend, begin block, deliver txs, end block, commit, repeat.

erikgrinaker:
Thanks Bez. I don't think benchmarks are necessary to merge this, especially since we don't really have a good general-purpose benchmarking suite at the moment. If you want to spend some time on it @mvdan then it's definitely welcome, but otherwise we can punt it and do some performance optimizations later if necessary.

mvdan:
How did you run those iavl tests, beyond replacing the tm-db module with my branch? I assume you had to make some change somewhere to use the badger backend, because I don't think iavl tests against all tm-db backends.

Also, your suspicion about that missing `iter.Valid` call was spot on - I added a couple of iterator test cases on an empty database, and got the same panic as you did:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x81d47d]

goroutine 90 [running]:
testing.tRunner.func1.1(0x92a2c0, 0xc8cc90)
	/home/mvdan/tip/src/testing/testing.go:1057 +0x30d
testing.tRunner.func1(0xc000082480)
	/home/mvdan/tip/src/testing/testing.go:1060 +0x41a
panic(0x92a2c0, 0xc8cc90)
	/home/mvdan/tip/src/runtime/panic.go:969 +0x175
github.com/dgraph-io/badger/v2.(*Item).Key(...)
	/home/mvdan/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.3/iterator.go:70
github.com/dgraph-io/badger/v2.(*Iterator).Item(...)
	/home/mvdan/go/pkg/mod/github.com/dgraph-io/badger/v2@v2.0.3/iterator.go:504
github.com/tendermint/tm-db.(*BadgerDB).iteratorOpts(0xc0000a00f8, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x1, 0x64, 0x1, ...)
	/home/mvdan/git/tm-db/badger_db.go:140 +0x23d
github.com/tendermint/tm-db.(*BadgerDB).ReverseIterator(0xc0000a00f8, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x10, 0xc00b2343c0, 0x0, ...)
	/home/mvdan/git/tm-db/badger_db.go:163 +0x106
github.com/tendermint/tm-db.testDBIterator(0xc000082480, 0x9a78c4, 0x8)
	/home/mvdan/git/tm-db/backend_test.go:314 +0x2b77
github.com/tendermint/tm-db.TestDBIterator.func1(0xc000082480)
	/home/mvdan/git/tm-db/backend_test.go:159 +0x45
testing.tRunner(0xc000082480, 0xc00555cc20)
	/home/mvdan/tip/src/testing/testing.go:1108 +0xef
created by testing.(*T).Run
	/home/mvdan/tip/src/testing/testing.go:1159 +0x386
exit status 2
```

Should be all good to go, I think, unless there are other iavl panics or errors.

mvdan:
Also, I've kept multiple commits, but I've also merged some changes together to not explode this into 15 commits. I think this is nicer as a series of commits than as a single commit conflating my work with Anton's.

erikgrinaker:
Thanks @mvdan! IAVL tests pass - to run them, I point `go.mod` to a local clone and manually update `TestRandomOperations` to use BadgerDB.

Could you update the `README` and `CHANGELOG` as well please?

> Also, I've kept multiple commits, but I've also merged some changes together to not explode this into 15 commits. I think this is nicer as a series of commits than as a single commit conflating my work with Anton's.

Thanks! We prefer to squash-merge PRs, since this makes it easier to e.g. cherry-pick or revert them.

mvdan:
> Could you update the `README` and `CHANGELOG` as well please?

Whoops, forgot about those. Done.

> Thanks! We prefer to squash-merge PRs, since this makes it easier to e.g. cherry-pick or revert them.

Of course - I would probably go for the same :) Do you want me to squash myself then, to provide a good commit message on the squashed commit? Or are you OK with a squashed commit message that simply lists all original commit messages?

erikgrinaker:
Thanks! I can squash-merge it.

PR #134, opened 2020-11-09
add Range to NewIterator

klim0v:
I see a slowdown when using iterations in the https://github.com/cosmos/iavl/pull/324#issuecomment-723961546 library in the case where the given key cannot be found. I guess this is because the `newGoLevelDBIterator` function uses `Seek()`, which has to go through the whole tree to make sure the key is missing.


klim0v:
> Error: requested golangci-lint version 'v1.26' isn't supported: we support only v1.28.3 and later versions

Can you fix the lint/golangci config?

marbar3778:
> Can you fix the lint/golangci config?

Could you make this PR against master, this error won't arise there, after that it can be back ported?

melekes:
wow, what a great find!

goleveldb.go
```diff
@@ -175,7 +176,7 @@ func (db *GoLevelDB) Iterator(start, end []byte) (Iterator, error) {
 	if (start != nil && len(start) == 0) || (end != nil && len(end) == 0) {
 		return nil, errKeyEmpty
 	}
-	itr := db.db.NewIterator(nil, nil)
+  itr := db.db.NewIterator(&util.Range{Start: start, Limit: end}, nil)
```
erikgrinaker, line 179:
```suggestion
	itr := db.db.NewIterator(&util.Range{Start: start, Limit: end}, nil)
```

erikgrinaker:
Can you add an entry for this in `CHANGELOG_PENDING.md` as well please?

goleveldb.go
klim0v: resolve

goleveldb.go
```diff
@@ -2,6 +2,7 @@ package db
 
 import (
 	"fmt"
+	"github.com/syndtr/goleveldb/leveldb/util"
 	"path/filepath"
 
 	"github.com/syndtr/goleveldb/leveldb"
```
erikgrinaker, line 8:
```suggestion
	"path/filepath"

	"github.com/syndtr/goleveldb/leveldb"
	"github.com/syndtr/goleveldb/leveldb/util"
```

klim0v:
> Can you add an entry for this in `CHANGELOG.md` as well please?

It's like that?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- [\#134](https://github.com/tendermint/tm-db/pull/134) Add the `util.Range` to `NewIterator` (@klim0v)
```
erikgrinaker, line 5:
```suggestion
- [\#134](https://github.com/tendermint/tm-db/pull/134) Improve GoLevelDB iterator performance by bounding underlying iterator range (@klim0v)
```

erikgrinaker (approved):

erikgrinaker:
Thanks! I'll run some tests with IAVL and Tendermint using this change, and if it looks good I'll backport to 0.5.x and cut new releases - won't be able to do this until tomorrow though, since I need to finish up some other work tonight.

klim0v:
Thank you, I was happy to contribute to the common cause!

PR #144, opened 2021-01-20
testing: docker deployment

marbar3778:
## Description

Add docker deployment for images used in this repo, and only here.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,48 @@
+name: Build & Push TM-DB-Testing
+on:
+  pull_request:
+    paths:
+      - "tools/*"
```
cmwaters, line 5: Is this right?

marbar3778: yup. I had to double-check myself because of the beautiful yaml/yml syntax.

cmwaters (approved):

PR #147, opened 2021-01-29
[WIP] transaction: transaction interface to solve atomicity issue in consensus/store.go

githubsands:
Includes transaction interface mentioned here https://github.com/tendermint/tendermint/issues/5888.

Currently is a WIP. Please let me know if you guys have any feed back- am new to this code base.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/tendermint/tm-db
 go 1.12
 
 require (
+	github.com/coreos/etcd v3.3.10+incompatible
 	github.com/dgraph-io/badger/v2 v2.2007.2
 	github.com/facebookgo/ensure v0.0.0-20160127193407-b4ab57deab51 // indirect
 	github.com/facebookgo/stack v0.0.0-20160209184415-751773369052 // indirect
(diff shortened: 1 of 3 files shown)


dshulyak:
i don't think that it is possible to stuff tx interface on top kv interface just like that, it will require significant amount of work.
tx usually implies at least atomicity and durability, sometimes particular isolation mode.

consider what state db will end up in if application performs 2 Set's in the same tx and machine goes away in the middle of tx? there is no guarantee whatsoever. this is exactly what transactions are about, proper transactions will ensure that either all changes will make it to the disk or none. just for this last bit you will need a proper recovery protocol (aries or similar), and there are other things that will require attention (what sort of concurrency mode is required by the app?).

i think what is expected is to provide generic interface, and implement it for every existing tm-db backend. tbh i don't know if every kv store has a transactional interface, at least you can fallback to a batch if it doesn't, for instance goleveldb enforces that only one transaction can be opened at a time, and most of the time people are trying to use batch for atomic updates if serializability is not a requirement.

i would recommend to fix that issue with a batch write, unless you are really committed to providing transactional interface for tm-db :). maybe @erikgrinaker will have different thoughts

erikgrinaker:
@dshulyak is right, transactions can't just be bolted on top of a kv database like this -- the idea was for this to expose and abstract over the underlying transaction functionality of the backend databases. However, as you point out, not all of the backends even support transactions, so this may require dropping backends.

The Tendermint team is considering simply dropping tm-db and picking a single database that provides the functionality that we need (e.g. ACID transactions) and making full use of it. In the meanwhile, batches are probably the best stop-gap solution.

githubsands:
Uses batches instead

PR #149, opened 2021-02-07
refactor: split backends into packages, add metadb

### LINKED ISSUE #113: current package layout hurts compile time and binary sizes
Since all the DB libraries are imported from a single package, using a single one of them requires building all of them. And since each is registered globally, they must be bundled in the final binary too.

This will get worse as more DB implementations are added (#38). I made a write-up about this, and potential solutions, on https://github.com/ChainSafe/chaindb/issues/2 since that module has a similar issue.

Similarly, I think a database implementation called `foo` should live at `github.com/tendermint/tm-db/foo`. Then, to use it, one would need to call a constructor like `foo.New(...)` instead of `db.NewDB("foo", ...)`. This way, because the user only imports the database implementations they need, Go only builds and links the libraries as needed, not all of them at once.

mvdan:
(see commit message, which is updated with pushes)


mvdan:
Alright, here's v1 of the huge refactor. One test is still being picky, I'll figure that out tomorrow. Otherwise this should all work - thoughts welcome.

It should go without saying that this is a big breaking change, but updating the users shouldn't be too hard. It should really be just replacing `db.NewDB(...)` calls with `metadb.NewDB(...)` to keep the old behavior with build tags.

internal/dbtest/dbtest.go
```diff
@@ -51,30 +47,71 @@ func checkItem(t *testing.T, itr Iterator, key []byte, value []byte) {
...
+				}
+				val >>= 6
+			}
+		}
+	}
+
+	return string(chars)
+}
+
+func Int642Bytes(i int64) []byte {
```
melekes, line 104: godoc is missing

internal/dbtest/dbtest.go
```diff
@@ -51,30 +47,71 @@ func checkItem(t *testing.T, itr Iterator, key []byte, value []byte) {
...
+	buf := make([]byte, 8)
+	binary.BigEndian.PutUint64(buf, uint64(i))
+	return buf
+}
+
+func Bytes2Int64(buf []byte) int64 {
```
melekes, line 110: godoc is missing

melekes (approved):
❤️ great work @mvdan

melekes:
Looks like some tests are failing:

```
# github.com/tendermint/tm-db/metadb [github.com/tendermint/tm-db/metadb.test]
metadb/db_cleveldb.go:14: undefined: metadb in metadb.RegisterDBCreator
metadb/db_cleveldb.go:14: undefined: clevelDBbCreator
FAIL	github.com/tendermint/tm-db/metadb [build failed]
# github.com/tendermint/tm-db/metadb
FAIL	github.com/tendermint/tm-db/remotedb [build failed]
metadb/db_cleveldb.go:14:15: undefined: metadb
metadb/db_cleveldb.go:14:57: undefined: clevelDBbCreator
FAIL	github.com/tendermint/tm-db/remotedb/grpcdb [build failed]
```

mvdan:
Yeah, I realise some godocs are missing and the tests aren't passing - I was hoping to get an early review before I spend more time polishing it. The overall refactor sounds good, then?

I don't think it's worth adding godocs to the internal test funcs, though. They didn't have it before, and they are still internal test APIs.

marbar3778:
Does it make sense to remove the name of the db from the file name? It's already in the respective folder. Badgerdb/badger_db.go could be badgerdb/db.go.

melekes:
> Does it make sense to remove the name of the db from the file name? It's already in the respective folder. Badgerdb/badger_db.go could be badgerdb/db.go.

Or it can be just `badger/db.go`.

melekes:
> The overall refactor sounds good, then?

yep

mvdan:
> Or it can be just `badger/db.go`.

I actually named the packages `foodb` instead of `foo` for a good reason; to keep the package names explicit. That is also why I rename `tm-db` to `tmdb` in the code instead of using just `db`, which is far too generic. Similarly, package names like `mem`, `meta`, `remote`, or `golevel` would be pretty confusing.

I agree with reducing repetition with filenames, though. I thought about doing the same with types, renaming `badgerdb.BadgerDB` to just `badgerdb.DB`, but I didn't want to refactor yet more things in this already-huge PR.

mvdan:
Alright, all tests should pass now. I've also updated CI and the makefile with the new memdb and goleveldb build tags.

mvdan:
I'm not sure what to do about the CI failures. The test failures are `signal: illegal instruction (core dumped)`, which I cannot reproduce. golangci-lint seems to fail at loading a package due to missing Cgo, which I also don't really understand as I don't use golangci-lint.

marbar3778:
It seems to the docker image. I was able to reproduce the issue with this command

```
test-all-docker:
	@echo "--> Running go test"
	@docker run --rm -v $(CURDIR):/workspace --workdir /workspace tendermintdev/docker-tm-db-testing go test $(PACKAGES) -tags memdb,goleveldb,cleveldb,boltdb,rocksdb,badgerdb -v
.PHONY: test-all-docker
```

mvdan:
I see. I can't fix that as I don't own the image. Maybe I should revert the CI changes and let one of you do that after this PR is merged. Or feel free to push follow-up commits on this same PR if you want to fix it here.

marbar3778:
The docker file is located in `/tools`, if you'd like to take a stab, otherwise reverting is fine. We can take a look later. The image will automatically be deployed since the file was touched. 

I would have to do some testing, I'm unsure why its failing.

marbar3778:
It's odd that I need to download rocksdb to run the linter.

marbar3778:
this PR fixes it: https://github.com/tendermint/tm-db/pull/151

not sure how linting fixed the tests.

marbar3778:
closing this in favor of #151

PR #151, opened 2021-02-11
metadb: split backends into packages, add metadb

### LINKED ISSUE #113: current package layout hurts compile time and binary sizes

marbar3778:
The old structure was a single huge package with all the DB backends in
it. Since tm-db users often want just one or two DB backends, and
building plus linking all of them is very heavy, build tags were used to
opt into some of the backends.

This same mechanism is now in metadb, for those who want it; it allows
opting into DB backends via build tags, without having to edit the code
directly.

Note that the old package included the memdb and goleveldb backends by
default. The new metadb does not include any by default, to make
backends explicit and more consistent. This required using more build
tags in CI.

Most users will want to statically use one DB implementation, without
metadb. Each separate sub-package does just that. They are direct moves
of the old code; the only major change is that constructors like
db.NewBadgerDB are now called badger.NewDB, to avoid the repetition of
badger.NewBadgerDB.

The root package error types are also exported now, to allow being used
by all separate packages. Note that this means they're now part of the
public API too; this should be fine.

Some test code also had to be shuffled around. All the tests which dealt
with backends were moved to metadb. The shared test helper code had to
be moved to an internal package, to remain reachable.

Finally, note that we consistently use "tmdb" as the imported package
name for the root package, since "db" is often used as a variable name.
This prevents confusion and common build errors.

Fixes #113.

Thank you @mvdan for the contribution.


melekes (approved):

marbar3778:
So the tests pass here but not on master, will dig into it

klim0v:
i think this change should have tag v0.7.0

marbar3778:
you are correct! ill see if the tag is still around

PR #187, opened 2021-09-11
Add an option to create MemDB iterators with no Mutex for the CacheKVStore

ValarDragon:
This is what we should be using in the SDK for the memDB CacheKVStore invocation there, to get the same guarantees around iterator closing as is present now. (Also this is just faster)

After this, I'll submit a PR that significantly improves performance of Next operation. (Removes lots of mutex ticking time)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 *.out
 
 .idea
+vendor/*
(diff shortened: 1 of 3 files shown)


marbar3778:
sorry, all looks goog

PR #197, opened 2021-11-08
testing: tools/Dockerfile: Upgrade RocksDB to 6.24.2

roysc:
Updates Rocksdb install to include a version with bindings needed for new `cosmos-sdk/db` module.
See https://github.com/cosmos/cosmos-sdk/pull/10470


marbar3778 (approved):

marbar3778:
@roysc can you fix conflicts here?

creachadair:
> @roysc can you fix conflicts here?

Got your back. 🙂

creachadair:
I believe the tests here (and on #194) will fail until the Docker workflow pushes a new version of the image. I _think_ that will happen automatically soon, but I'm not positive. That image hasn't been updated in a while, and the credentials could be stale.

marbar3778:
I dont have permission to see what is in the secrets so i guess we merge and hop it works or someone who has permission looks into it

roysc:
The only change is the url and file name and I tested building the image locally, so it should be fine. But the illegal instruction errors are kind of weird, is it some architecture mismatch in the CI build? anyway, it is happening on all open PRs from what I see.

creachadair:
> The only change is the url and file name and I tested building the image locally, so it should be fine. But the illegal instruction errors are kind of weird, is it some architecture mismatch in the CI build? anyway, it is happening on all open PRs from what I see.

Yes, the image hadn't been updated in a while, so in addition to the Go version change I suspect there may have been ABI or OS version changes. The new tag should help, I think.

creachadair:
Something is weird in the build. I'll figure out whatever it is, and merge this once it's working again.

creachadair:
Ok, I sorted out the build issues, but I believe in the current state, the updated rocksdb version is not compatible with the C API expected by the Go library. I didn't attempt to fix that, but at least now the tests should be diagnostic.

creachadair:
> Ok, I sorted out the build issues, but I believe in the current state, the updated rocksdb version is not compatible with the C API expected by the Go library. I didn't attempt to fix that, but at least now the tests should be diagnostic.

Or, maybe I'm wrong. I had build failures with this version previously, but the don't seem to be manifesting here. Let's see what happens.

creachadair:
> Or, maybe I'm wrong. I had build failures with this version previously, but the don't seem to be manifesting here. Let's see what happens.

Wait, I see why: The tests are running against the _old_ version, because the build image does not update until _after_ the Dockerfile change is merged. I believe the issue is still real.

creachadair (changes requested):
We should make sure the tests pass in a container built from this config change. The one used for this PR is from the _last_ change, and doesn't reflect this update.

roysc:
My mistake, we also needed to update `gorocksdb` to the fork. The latest commit should pass.

marbar3778:
> Something is weird in the build. I'll figure out whatever it is, and merge this once it's working again.

secrets arent set for a contribution from a fork. we have to open this ourself to get it to pass

marbar3778:
Everything passes now. This should be good to merge. Ill wait for @creachadair to approve.

creachadair (approved):
It seems a little weird to introduce a fork that wasn't there before, but I guess we at least know this has been working in the SDK.

roysc:
Yeah, the `gorocksdb` maintainers were unresponsive about merging bindings we need, so I had to fork the repo.

PR #203, opened 2021-11-08
Prepare a new compatibility release v0.6.6.

creachadair:
This PR prepares the changelog and commit history for a v0.6.6 release.

The API changes in tag v0.6.5 have been preserved in the new `package-reorg` branch, so that any existing dependents will continue to work. This change reverts the breaking API and package structure changes on master, and prepares to tag a new release compatible with the previous v0.6.4 so that normal upgrades with `go get` will function correctly.

Other changes since v0.6.4, including config file changes and CI updates, are preserved.

Note that if we do want to tag a new release with the breaking API changes, it will need to use a different import path to avoid colliding with the semver namespace of the main branch. This PR does not address that, but keeps the work around so we can handle that separately. A submodule would be one option, if we don't want to split the repository.


cmwaters (approved):
I haven't scrutinised all these changes as I assume they all come from the reversion so it LGTM 👍 . I would advocate, as I don't think it will be too much work, to *eventually* dump all the breaking API changes into a `v2` directory and cut it as a v2 go module. I'm fine with having this as a 1.0 and 2.0 as I don't foresee too much work happening with this repo other than the odd version bump and thus these API's should be pretty stable.

marbar3778 (changes requested):
please work off of https://github.com/tendermint/tm-db/tree/release/0.6.x for a 0.6.x release. No need to revert code.

tychoish (approved):

creachadair:
> please work off of https://github.com/tendermint/tm-db/tree/release/0.6.x for a 0.6.x release. Not sure theres a need to make this more complicated

I have done the rebase, but we can't eliminate the revert, since the API changes are already on the branch history. Well, we could rebase them away, but then the history would diverge not only from the 0.6.5 release but also master. If you prefer that we can do it, though.

marbar3778 (approved):
sorry for delay. I meant to remove my change request since we will be deprecating this library anyways

marbar3778:
do you think we could get https://github.com/tendermint/tm-db/pull/205 before doing 0.6.6?

creachadair:
> sorry for delay. I meant to remove my change request since we will be deprecating this library anyways

I merged it into this PR from the side-spur of the release branch. I think it probably _does_ make sense to include it in v0.6.6, and I updated the changelog on that basis.

PR #204, opened 2021-11-09
 dockerfile update rocksdb

marbar3778:


creachadair:
Ah, this explains why my tests suddenly started failing.

marbar3778:
oh my bad. I need to version the image.

creachadair:
> oh my bad. I need to version the image.

Probably a better solution is to build the image as part of the workflow. Then there's no need to track yet another version stamp, and we can probably get GitHub to cache the layers for us.

PR #205, opened 2021-11-12
Add an option to create MemDB iterators with no Mutex for the CacheKV…

marbar3778:
…Store (#187)

* Add an option to create MemDB iterators with no Mutex for the CacheKVStore

* Update changelog

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 *.out
 
 .idea
+vendor/*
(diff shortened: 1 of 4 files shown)


creachadair:
I merged this in with #203, so it is now included in v0.6.6.

marbar3778:
thank you!

PR #207, opened 2021-12-03
Update gorocksdb replace directive to use cosmos fork

roysc:
Change go.mod to use the new fork under cosmos: https://github.com/cosmos/gorocksdb


creachadair (approved):

Karam19925 (commented):
> #207 reviewed

PR #214, opened 2022-01-25
cosmos/gorocksdb in the imports

faddat:
* updates .golangci-lint.yml
* uses cosmos/gorocksdb everywhere preventing replace statements downstream
* removes the version number from golanci-lint so that we're always using the latest version (earlier one was dated and it couldn't run because it didn't recognize `revive`)

### DIFF
--- a/common_test.go
+++ b/common_test.go
@@ -127,7 +127,7 @@ func benchmarkRandomReadsWrites(b *testing.B, db DB) {
 			val := internal[idx]
 			idxBytes := int642Bytes(idx)
 			valBytes := int642Bytes(val)
-			//fmt.Printf("Set %X -> %X\n", idxBytes, valBytes)
+			// fmt.Printf("Set %X -> %X\n", idxBytes, valBytes)
 			err := db.Set(idxBytes, valBytes)
 			if err != nil {
 				// require.NoError() is very expensive (according to profiler), so check manually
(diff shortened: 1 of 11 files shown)


marbar3778 (approved):

common_test.go
```diff
@@ -127,7 +127,7 @@ func benchmarkRandomReadsWrites(b *testing.B, db DB) {
 			val := internal[idx]
 			idxBytes := int642Bytes(idx)
 			valBytes := int642Bytes(val)
-			//fmt.Printf("Set %X -> %X\n", idxBytes, valBytes)
+			// fmt.Printf("Set %X -> %X\n", idxBytes, valBytes)
```
creachadair, line 130: We might as well just delete it.
```suggestion
```

common_test.go
```diff
@@ -145,7 +145,7 @@ func benchmarkRandomReadsWrites(b *testing.B, db DB) {
 				// require.NoError() is very expensive (according to profiler), so check manually
 				b.Fatal(b, err)
 			}
-			//fmt.Printf("Get %X -> %X\n", idxBytes, valBytes)
+			// fmt.Printf("Get %X -> %X\n", idxBytes, valBytes)
```
creachadair, line 148: ibid.
```suggestion
```

creachadair (approved):

faddat:
makes sense and thanks!

PR #216, opened 2022-02-07
goleveldb: ForceCompact

marbar3778:


creachadair (commented):
The change seems fine, but it would be helpful if the description gave a bit of context for why this is necessary.

marbar3778:
yes, in the sdk and tendermint we Call delete on the db but with lam trees the space on disk is not reclaimed until compaction is completed, in our environment we write so often the db doesn't have enough time to sufficiently compact itself. By adding force compact I have two things planned, in the short term I wrote a tool that prunes blocks, state and app state then calls force compact on the entire db. With this tool, after pruning the stores I would need to compact it. In the mid term id like to add force compacting to the sdk to avoid seekcompaction on goleveldb. 

The overall goal is optimisations with this. It allows the application to control compaction instead of passing it on to the default settings of the db

creachadair (approved):

PR #218, opened 2022-02-15
Update rocksdb.go

faddat:
Today I'm working to make rocks an acceptable default, and this seems to help.  I got some of this from terra's work on tm-db.

If desired, I could merge my version of this with more options illustrated for the user, but commented out:

https://github.com/notional-labs/tm-db/blob/rocks-tuning/rocksdb.go

I should also highlight Terra's work here:

https://github.com/tendermint/tm-db/compare/master...terra-money:performance

And finally I should mention that a company that specalizes in rocksdb performance enhancements has reached out to me.  I bungled scheduling last week, reckon they will end up reading this PR, and hope to succeed with scheduling this week. 

I figure it's helpful for me to show exactly what I am working on and with-- basically, Juno has has a longstanding sync issue.  So I am trying to make it possible to do a pure-rocks node with Osmosis' iavl updates.  This way, it'll be as easy as possible for people to set up archives.  

I'm then going to go through every version of juno and apply the changes.  So far, Terra's work seems to perform best. 

Currently I am always talking in generalities.  I am not sure what the best way to look at this stuff is:

* blocks per second synced?
* queries per second against a fully synced archive node without making the node fall behind the tip?
* something else?

Happy to compare notes with anyone. 

Furthermore, it's notable that this will immediately apply to any 44 series chain, including osmosis and gaia.

### DIFF
--- a/rocksdb.go
+++ b/rocksdb.go
@@ -37,6 +37,8 @@ func NewRocksDB(name string, dir string) (*RocksDB, error) {
 
 	opts := gorocksdb.NewDefaultOptions()
 	opts.SetBlockBasedTableFactory(bbto)
+	// SetMaxOpenFiles to 4096 seems to provide a reliable performance boost
+	opts.SetMaxOpenFiles(4096)
 	opts.SetCreateIfMissing(true)
 	opts.IncreaseParallelism(runtime.NumCPU())
 	// 1.5GB maximum memory use for writebuffer.


marbar3778 (approved):
is there a benchmark to prove the improvement?

faddat:
@marbar3778 in fact, @ValarDragon tells me that this isn't necessary to use fast iavl, and that the osmo import simply wasn't updated.  I am going to close this.

In terms of benchmark, not yet.  

Do you have any suggestions on benchmarking the performance of mainnet nodes?

ValarDragon (approved):
LGTM!

The default is 1000 files (defined here: https://github.com/tecbot/gorocksdb/blob/master/options.go#L373-L380 )

Reading the comments there, it appears to me that this does increase the max RAM usage from 2GB to 8GB then. (Which doesn't sound crazy, but maybe indicative of something that should eventually go into a config file)

creachadair:
Do we need to be concerned about expanding existing deployments' RAM footprint by a factor of approximately 4, if they were to patch this update? It seems like a lot of deployments depend on container orchestration. How does the operator know they need to bump up their RAM allocation (and possibly their instance sizes)?

ValarDragon:
RocksDB usage right now is solely experimental AFAIK, seems like a thing that can be communicated with a warning due to it not being actively used as of yet

creachadair:
I guess we'll see how it shakes out. 🙂

PR #219, opened 2022-02-21
update 0.6.x with master

marbar3778:


marbar3778:
ill use the head of master to release instead

PR #229, opened 2022-04-11
gorocksdb -> grocksdb & rocksdb v7.0.3

faddat:
This PR changes the wrapper that we use for rocksdb to one that keeps up to date with facebook/rocksdb, so that we can support rocksdb v7 and test that.

I don't know exactly where to put this, but we are also working on pebbledb internally. 

This will fail CI until merged, when the Docker image can be updated.


faddat:
Closing to clean up.

PR #230, opened 2022-04-12
PebbleDB

faddat:
I spoke with the Terra team, and they mentioned that because their CosmWasm usage is high, there were issues with cgo/FFI (Rocks will hit this, as will CosmWasm, and this resulted in high overhead) when using RockDB, making it a less than ideal solution for their needs.  Since CosmWasm is being adopted pretty widely in Cosmos, we're experimenting with PebbleDB, which outperforms RocksDB in synthetic benchmark tests.

Currently, the goal of these changes is to be able to test them and see if they increase RPC capacity on relayers, so that the IBC relay network can be more reliable.


faddat:
Remaining things to do here include adding pebbledb to the iavl benchmarks

faddat:
this contains a bug affecting goleveldb that we added while messing about.  I'm going to close it and submit a pr that only touches pebble stuff.

PR #232, opened 2022-04-15
update goleveldb and badgerdb for benchmark purposes

faddat:
This PR brings goleveldb and badgerdb current to enable better comparisons


go.mod
```diff
@@ -4,13 +4,13 @@ go 1.17
 
 require (
 	github.com/cosmos/gorocksdb v1.2.0
-	github.com/dgraph-io/badger/v2 v2.2007.2
+	github.com/dgraph-io/badger/v2 v2.0.1-rc1.0.20210105060133-d6666aecfdc3
```
creachadair, line 7: This seems a bit sus—why are we targeting a release candidate?

It's fine to use whatever versions for benchmarking, but I think for a library shared by a lot of consumers, we ought to be conservative about what we point to.

(The RocksDB thing is a notable hole in that, but I don't think we should ream that hole out any further)

faddat: you're totally right about that, and even the benchmarks should be on released versions

faddat:
I need to reduce the scope of this pr and will do so tomorrow

PR #236, opened 2022-04-25
Upgrades gorocksdb to grocksdb

faddat:
If these are unwanted changes, please let me know, and I'll make a PR that removes RocksDB

NB: tests are failing because the docker image is out of date, specifically this PR makes the dockerfile use rocks v7, but the one that is currently used to run tests is rocks v6


faddat:
NB: This will continue to fail, as will tests in the cosmos-sdk, because our Dockerfile uses rocksdb v6.

Because of how CI is set up, this isn't avoidable, this PR would need to be merged.

Issue #156, opened 2021-03-24
Could we remove `mtx` from `prefixdb`?


jinsankim:
I'm investigating how to increase concurrency in terms of cosmos-sdk app. To my understanding, almost db access paths from `cosmos store` to `db backend` are serialized by `mtx`. I think it's not a good idea for performance. As a bottom layer of db access path, I'd like to remove `mtx` from `prefixdb`. (Please note that `cosmos/iavlStore` depends on this `prefixdb`).  `prefixdb` wraps another `tm-db` instance and this another `tm-db` is already concurrency safe by CONTRACT.  So, intuitionally, I think we could remove `mtx` from `prefixdb`. HDYT?

marbar3778:
Queries in the SDK are routed through Tendermint which holds a global mutex across the ABCI connection, this is the main bottleneck of the SDK outside the underlying storage issues with IAVL. In the near future the SDK and Tendermint will stop using this repo and house their own db interface implementations internally. This will lend itself to better performance. We are also working on making SDK state access concurrent with the recent storage ADR. 

This all being the case, I'm not sure if it is worth doing this work as it is still unclear if we will do another release with this repo.

jinsankim:
Even though this repo might be not used in the future, I'd like to discuss why `mtx` is needed for `prefixdb` and whether it's possible to remove it. Please could you share your thought?

marbar3778:
Quickly reading the code, the reason we would need a mutex is that we can't assume the underlying database is concurrently safe. This is an overall problem with this repo. Abstracting DBs means we can make fewer assumptions of the underlying db.

jinsankim:
AFAIK, the all db implementation should be concurrency safe. The description also shows it.

https://github.com/tendermint/tm-db/blob/53ed3dbaa3eb4c7693ca0aaa2deff093e4675b24/types.go#L16

ilovers:
hi @jinsankim , I have the same issue now. Did you make any progress afterwards?

jinsankim:
@ilovers https://github.com/line/tm-db/pull/20

marbar3778:
> @ilovers [line/tm-db#20](https://github.com/line/tm-db/pull/20)

Hey seems this change would be acceptable. Would you want to see about upstreaming it?

PR #239, opened 2022-04-25
fix: remove mutex in prefixdb

faddat:
closes:

https://github.com/tendermint/tm-db/issues/156

* Adds additional testing to all databases
* Adds .gitpod.yml so that devs can click a single button and ensure that changes work, in the exact docker environment that we use in ci


prefixdb.go
```diff
@@ -62,8 +62,8 @@ func (pdb *PrefixDB) Set(key []byte, value []byte) error {
 	if value == nil {
 		return errValueNil
 	}
-	pdb.mtx.Lock()
-	defer pdb.mtx.Unlock()
+	// pdb.mtx.Lock()
```
creachadair, line 65: If we're going to remove the mutex, we should delete it, not just comment it out.

creachadair (commented):
In light of the argument in #156, this seems like it should be safe. 

I am a bit worried that we don't seem to have any tests that exercise the documented concurrency-safety requirement, though -- and there is at least one case where the implementations differ about what operations are allowed to overlap (cf. https://github.com/tendermint/tm-db/blob/master/util_test.go#L29).

Running the race detector can't really help, when there is no concurrency for it to check. This PR doesn't create that problem, but it does make it more relevant.

prefixdb.go
faddat: you got it.

faddat:
@creachadair 

I may need to nuke-reinstall my test machine, but I did a bit of a survey on databases recently.

That led me to the terra-money/tmdb:performance branch.

The speed is super impressive but I've had an explosion (concurrent map read and write error).

Can you review my method for testing this stuff?

I am thinking like:

* make a gitlab instance
* automate a replace of tm-db
* run syncs of gaia and osmosis in parallel, check the impact on sync time
* when we've got synced nodes, do an automated test of queries with vegeta

faddat:
https://github.com/cosmos/gaia/pull/1415

marbar3778 (approved):

creachadair:
> Can you review my method for testing this stuff?
> 
> I am thinking like:
> 
> * make a gitlab instance
> * automate a replace of tm-db
> * run syncs of gaia and osmosis in parallel, check the impact on sync time
> * when we've got synced nodes, do an automated test of queries with vegeta

I'm not convinced we need anything that elaborate. Even just a regular Go test that creates a database instance and does some amount of parallel insertions, lookups, and deletions would give the race detector something to push against. I don't think we need an elaborate multi-process test harness to verify the basic concurrency properties the library is intended to guarantee.

Of course, integration tests are nice too, but as far as tm-db itself, integration is a concern for the clients of the library.

Very roughly, I was thinking along these lines:

```go
// Run generates concurrent reads and writes to db so the race detector can
// verify concurrent operations are properly synchronized.
// The contents of db are garbage after Run returns.
func Run(t *testing.T, ctx context.Context, db tmdb.DB) {
	t.Helper()

	const numWorkers = 10
	const numKeys = 32

	var wg sync.WaitGroup
	for i := 0; i < numWorkers; i++ {
		wg.Add(1)
		i := i
		go func() {
			defer wg.Done()

			// Insert a bunch of keys with random data.
			for k := 1; k <= numKeys; k++ {
				key := taskKey(i, k) // say, "task-<i>-key-<k>"
				value := someRandomValue()
				if err := db.Set(key, value); err != nil {
					t.Errorf("Task %d: db.Set(%q=%q) failed: %v", 
						i, string(key), string(value), err)
				}
			}

			// Iterate over the database to make sure our keys are there.
			it, err := db.Iterator(nil, nil)
			if err != nil {
				t.Errorf("Iterator[%d]: %v", i, err)
				return
			}
			found := make(map[string][]byte)
			mine := []byte(fmt.Sprintf("task-%d-", i))
			for it.Valid() {
				it.Next()
				if key := it.Key(); bytes.HasPrefix(key, mine) {
					found[string(key)] = it.Value()
				}
			}
			if err := it.Error(); err != nil {
				t.Errorf("Iterator[%d] reported error: %v" i, err)
			}
			if err := it.Close(); err != nil {
				t.Errorf("Close iterator[%d]: %v", i, err)
			}
			if len(mine) != numKeys {
				t.Errorf("Task %d: found %d keys, wanted %d", i, len(mine), numKeys)
			}

			// Delete all the keys we inserted.
			for key := range mine {
				if err := db.Delete([]byte(key)); err != nil {
					t.Errorf("Delete %q: %v", key, err)
				}
			}
		}()
	}
	wg.Wait()
}
```

N.B. this is not tested, but should be the right general flavour.

faddat:
@creachadair thank you!  it might even help me with the PebbleDB implementation

prefixdb_test.go
```diff
@@ -23,6 +25,28 @@ func mockDBWithStuff(t *testing.T) DB {
 	return db
 }
 
+func taskKey(i, k uint32) []byte {
+
```
creachadair, line 29: This could probably more simply be

```go
func taskKey(i, k int) []byte { return []byte(fmt.Sprintf("task-%d-key-%d", i, k)) }
```

We don't really care about compactness here, only uniqueness.

vuong177: Thanks for suggestion! I changed it in newest commit.

prefixdb.go
```diff
@@ -27,8 +27,8 @@ func (pdb *PrefixDB) Get(key []byte) ([]byte, error) {
 	if len(key) == 0 {
 		return nil, errKeyEmpty
 	}
-	pdb.mtx.Lock()
-	defer pdb.mtx.Unlock()
+	// pdb.mtx.Lock()
+	// defer pdb.mtx.Unlock()
```
creachadair, line 31:
```suggestion
```

prefixdb_test.go
```diff
@@ -20,6 +25,86 @@ func mockDBWithStuff(t *testing.T) DB {
...
+	return b
+}
+
+// Run generates concurrent reads and writes to db so the race detector can
+// verify concurrent operations are properly synchronized.
+// The contents of db are garbage after Run returns.
+func Run(t *testing.T) {
+
+	db := mockDBWithStuff(t)
```
creachadair, line 48: It won't help if we use a mock database here—this needs to exercise the real database implementation. I suggest making this a parameter, e.g.,

```go
func Run(t *testing.T, db DB) {
  // ...
}
```

Then we can use it to write more specific tests against the individual implementations, e.g.,

```go
func TestBadger(t *testing.T) {
   db := newTestBadgerDB(t) // or whatever we do to set one up
   t.Run("Badger", func(t *testing.T) { Run(t, db) })
}
```

or words to that effect. Exercising the MemDB is fine, but doesn't tell us much about the concurrency safety of the "real" backends.

faddat: gotcha!

OK, will address this tomorrow, and thanks!

catShaark: should we test prefix_db with many db implementations, like some kind of table-driven test

faddat: I do think that is what he means <3

prefixdb_test.go
```diff
@@ -49,7 +49,63 @@ func TestWithGolevelDB(t *testing.T) {
 
 	pdb := NewPrefixDB(db, bz("key"))
 
-	t.Run("PrefixDB", func(t *testing.T) { Run(t, pdb) })
+	t.Run("GolevelDB", func(t *testing.T) { Run(t, pdb) })
+}
+
+func TestWithClevelDB(t *testing.T) {
+	name := fmt.Sprintf("test_%x", randStr(12))
+	defer cleanupDBDir("", name)
```
creachadair, line 57: We can lean on the test harness to create and clean up test directories for us:

```suggestion
	dir := t.TempDir()
	path := filepath.Join(dir, "cleveldb")
	db, err := NewCLevelDB(path, "")
	require.NoError(t, err)
```

The test will auto-cleanup the directory. We might want to be sure it's closed first, though:

```go
defer func() { assert.NoError(t, db.Close()) }()
```

or words to that effect.

prefixdb_test.go
```diff
@@ -49,7 +49,63 @@ func TestWithGolevelDB(t *testing.T) {
...
+
+	db, err := NewCLevelDB(name, "")
+	require.Nil(t, err)
+
+	pdb := NewPrefixDB(db, bz("key"))
```
creachadair, line 62: I don't know that we need to test the prefixdb for _every_ backend. What I had in mind here was ensuring we had concurrency tests for all the backends, and then for the prefixed wrapper. (This isn't bad either, but it exercises two things at once)

faddat:
I've added .gitpod.yml to this

@creachadair if you wouldn't mind trying the gitpod browser extension, I think you'd instantly understand why it is a nice-to-have.

faddat:
That was dramatically easier than making a dev env, and now it's all set :D!

faddat:
I think this one is set now.

cleveldb_test.go
```diff
@@ -36,9 +38,9 @@ func BenchmarkRandomReadsWrites2(b *testing.B) {
 			idx := (int64(rand.Int()) % numItems)
 			internal[idx]++
 			val := internal[idx]
-			idxBytes := int642Bytes(int64(idx))
-			valBytes := int642Bytes(int64(val))
-			//fmt.Printf("Set %X -> %X\n", idxBytes, valBytes)
+			idxBytes := int642Bytes(idx)
+			valBytes := int642Bytes(val)
+			// fmt.Printf("Set %X -> %X\n", idxBytes, valBytes)
```
creachadair, line 43:
```suggestion
```

prefixdb_test.go
```diff
@@ -20,6 +26,143 @@ func mockDBWithStuff(t *testing.T) DB {
...
+func randomValue() []byte {
+	b := make([]byte, 16)
+	_, err := rand.Read(b)
+
+	if err != nil {
+		return nil
+	}
+
+	return b
+}
```
creachadair, line 42: We can simplify this as [rand.Read](https://pkg.go.dev/math/rand#Read) never reports an error:

```suggestion
func randomValue() []byte {
	b := make([]byte, 16)
	rand.Read(b)
	return b
}
```

prefixdb_test.go
```diff
@@ -20,6 +26,143 @@ func mockDBWithStuff(t *testing.T) DB {
...
+
+func TestWithGolevelDB(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, "goleveldb")
+
+	db, err := NewGoLevelDB(path, "")
+	require.NoError(t, err)
+
+	t.Run("GolevelDB", func(t *testing.T) { Run(t, db) })
+}
```
creachadair, line 52: Since these are split out by backend, we can skip the additional indirection through the subtests in each of these (the top-level test function will already be named in the output anyway), e.g.:
```suggestion
func TestGolevelDB(t *testing.T) {
	path := filepath.Join(t.TempDir(), "goleveldb")

	db, err := NewGoLevelDB(path, "")
	require.NoError(t, err)

	Run(t, db)
}
```

.gitpod.yml
```diff
@@ -0,0 +1,7 @@
+# This configuration file was automatically generated by Gitpod.
```
creachadair, line 1: This file isn't related to this change, let's leave it out of the PR.

prefixdb_test.go
```diff
@@ -1,6 +1,12 @@
 package db
```
creachadair, line 1: As written, this test can't run unless all the build tags are enabled. I think we should split out the tests that require specific build tags (e.g., CLevelDB, RocksDB, BadgerDB, BoltDB) into tagged files so they only run with those tags enabled. Otherwise even if someone doesn't want to use those backends, they are forced to install all of the support libraries just to run the unit tests.

creachadair (dismissed):
This looks pretty close, there are just a few details to sort out.

.gitpod.yml
faddat: great idea, thank you

prefixdb_test.go
faddat: OK, I thought about doing that, can do that now...

faddat:
@creachadair I hope to have the tests moved today, wish me luck...

faddat:
is this wanted?

marbar3778:
id vote for yes

marbar3778 (approved):

marbar3778:
generally the scope should be fixed on a single item in prs. Expanding scope makes it take longer to get things merged.

PR #241, opened 2022-05-01
Badgerdb  Settings

faddat:
- update badgerdb
- vscode so that gopls works for everyone
- give badger a cache and have it stop doing things that should never happen because tm/iavl would handle them


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
```
creachadair, line 1: I'd prefer not to check this sort of thing into the repo. People can drop their own VSCode settings in there if they want, there's nothing complex about this project.

(Though I wouldn't object to adding this to the `.gitignore` so that adding the file won't leave an untracked entry in the working dir)

As for the build tags: Even if we _do_ add editor configs, I'd argue we shouldn't make the default something that will require the user to install additional software to build and run tests. Both `cleveldb` and `rocksdb` require additional installations locally. I'd suggest folks should have to "opt in" to tags requiring external libraries.

badger_db.go
```diff
@@ -28,7 +28,14 @@ func NewBadgerDB(dbName, dir string) (*BadgerDB, error) {
 	if err := os.MkdirAll(path, 0755); err != nil {
 		return nil, err
 	}
+
+	valueThreshold := int64(5 << 20) // 5MB
```
creachadair, line 32: I'm not sure where these values came from.  

Are these numbers chosen to address some specific concern?  If so: Is there an issue we could link to understand why these particular values are good enough when the defaults aren't?

I'd prefer to be cautious about overriding the defaults, particularly for a library used on potentially a lot of different systems. Increasing size limits and concurrency can have implications for containerized deployments that have to tune their RAM budget, descriptor limits, etc.

Re: DetectConflicts: Disabling the conflict checking _might_ be safe in some cases, but it does affect the semantics of the implementation. It'd be good to have some tests to exercise concurrent Get/Set/Delete workloads to have confidence this doesn't allow inconsistent views.

faddat: These... are bad. I'll update the branch, with finalized code :)

faddat: The value threshold comes from this:

https://github.com/tharsis/evmos/pull/542

I also want to throw a pretty huge red flag on badger:

https://github.com/dgraph-io/badger/pull/1754

If that PR is merged, then we can use badger, and fi not, pls see 542 on evmos-- large genesis.json files will fail.

faddat:
This is good to go now.  The changes you question have to do with a lack of configuration options in badgerdb, which I do recommend we look into before adopting it as canonical.

faddat:
https://github.com/dgraph-io/badger/pull/1754

badger_db.go
creachadair: FWIW, Tendermint since v0.35 (specifically after https://github.com/tendermint/tendermint/pull/6554) does not store the genesis document in the database anyway (regardless which backend is used).

I expect applications will have different storage profiles than the core consensus node, and that's fine. One of the goals of decoupling the consensus storage from application storage is to allow those two profiles to be different.

badger_db.go
```diff
@@ -28,7 +28,12 @@ func NewBadgerDB(dbName, dir string) (*BadgerDB, error) {
 	if err := os.MkdirAll(path, 0755); err != nil {
 		return nil, err
 	}
+
+
 	opts := badger.DefaultOptions(path)
+	opts.IndexCacheSize = 256 << 20 // 100 mb or some other size based on the amount of data
```
creachadair, line 34: I'm fine with the idea of setting some reasonable baseline value here, but we should document where it came from.

According to Badger's docs, the Bloom filter for each SSTable is 5MB (presumably MiB). Do we have any data about typical table counts for production networks?  (We don't need to tune this super precisely, but the reader ought to have _some_ idea of where we got the value)

badger_db.go
```diff
@@ -28,7 +28,12 @@ func NewBadgerDB(dbName, dir string) (*BadgerDB, error) {
...
+	opts.DetectConflicts = false
```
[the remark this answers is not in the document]
faddat: hmmmmmm

OK, I will remove that setting, thank you!

badger_db.go
faddat: I don't, but I like where this is heading and I'll get that in here, too.

badger_db.go
```diff
@@ -28,7 +28,12 @@ func NewBadgerDB(dbName, dir string) (*BadgerDB, error) {
...
+	opts.NumGoroutines = 64
```
faddat, line 36: @creachadair this was a guess.

Supposing that we're rejecting any test that doesn't occur against a live node, do you have any thoughts on getting data behind this?

creachadair: It doesn't necessarily have to be a live node, I just think we should have some reason for choosing a value besides a general feeling 🙂 

Even a load test could be helpful, or a testnet that gets enough load for the difference to matter.

PR #245, opened 2022-05-06
Rocksdb: create folder if it doesn't exist

faddat:
Instead of panicking and crashing, if rocksdb finds that it doesn't have a folder (eg: snapshots.db) it will now make it.

Tested and working here:

https://github.com/tharsis/evmos/pull/542


creachadair (commented):
The description says rocks, but touches badger also, and a bunch of other unrelated changes. Let's please keep the cleanup changes separate from the changes that do more meaningful stuff.

Note that BadgerDB doesn't require this treatment: It will create the prefix of the path it's given if necessary. (I don't know about Rocks)

faddat:
oh sure thing!

I have a habit of always fumpting.  1 moment please

rocksdb.go
```diff
@@ -50,7 +51,10 @@ func NewRocksDBWithOptions(name string, dir string, opts *gorocksdb.Options) (*R
 	dbPath := filepath.Join(dir, name+".db")
 	db, err := gorocksdb.OpenDb(opts, dbPath)
 	if err != nil {
-		return nil, err
+		err = os.MkdirAll(dbPath, 0755)
```
creachadair, line 54: `gorocksdb.OpenDb` calls the C library's `rocksdb_open` which already creates the directory if it doesn't exist ([ref](https://github.com/facebook/rocksdb/blob/736a7b5433025a84efc01836c145ee5afa5cffb1/db/db_impl/db_impl_open.cc#L1685)). So this is harmless, but should not be necessary for RocksDB.

faddat: not for state sync snapshots.

I am not sure why that is the case, but it is the case.

creachadair: I suppose there is no harm in creating the directory. But isn't this backwards? If we don't create the directory till after the `OpenDb` call has already failed, the code below will fail anyway.

I'm guessing this probably wants to be something like

```go
if err := os.MkdirAll(dbPath, 0700); err != nil {
   return nil, fmt.Errorf("creating database path: %w", err)
}
db, err := gorocksdb.OpenDb(opts, dbPath)
if err != nil {
   return nil, fmt.Errorf("opening database: %w", err)
}
```

or words to that effect.

faddat: I don't know why this works.  Let me show ya what I have......

PR #246, opened 2022-05-06
Boltdb: Make folder if it doesn't exist

faddat:
This allows boltdb to make a folder if it doesn't find one instead of panicking and crashing.

### DIFF
--- a/boltdb.go
+++ b/boltdb.go
@@ -49,7 +49,7 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
 	if err := os.MkdirAll(filepath.Dir(dbPath), 0o0700); err != nil {
 		return nil, err
 	}
-	dblocation := filepath.Join(dbpath, "boltdb")
+	dblocation := filepath.Join(dbPath, "boltdb")
 	db, err := bbolt.Open(dblocation, os.ModePerm, opts)
 	if err != nil {
 		return nil, err
--- a/boltdb.go
+++ b/boltdb.go
@@ -46,12 +46,12 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
 	}
 
 	dbPath := filepath.Join(dir, name+".db")
-	if err := os.MkdirAll(filepath.Dir(dbPath), 00700); err != nil {
-   	return nil, err
+	if err := os.MkdirAll(filepath.Dir(dbPath), 0o0700); err != nil {
+		return nil, err
 	}
 	db, err := bbolt.Open(dbPath, os.ModePerm, opts)
 	if err != nil {
-	return nil, err
+		return nil, err
 	}
 
 	// create a global bucket
--- a/boltdb.go
+++ b/boltdb.go
@@ -46,12 +46,12 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
 	}
 
 	dbPath := filepath.Join(dir, name+".db")
+	if err := os.MkdirAll(filepath.Dir(dbPath), 00700); err != nil {
+   	return nil, err
+	}
 	db, err := bbolt.Open(dbPath, os.ModePerm, opts)
 	if err != nil {
-		err = os.MkdirAll(dbPath, 0o755)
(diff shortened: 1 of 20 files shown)


boltdb.go
```diff
@@ -49,8 +49,11 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
 	dbPath := filepath.Join(dir, name+".db")
 	db, err := bbolt.Open(dbPath, os.ModePerm, opts)
 	if err != nil {
-		return nil, err
-	}
+		err = os.MkdirAll(dbPath, 0755)
```
creachadair: The target still needs to be a file, so if we create the whole path it will still fail. What I think we probably want here is something like:

```go
if err := os.MkdirAll(filepath.Dir(dbPath), 00700); err != nil {
   return nil, err
}
db, err := bbolt.Open(dbPath, os.ModePerm, opts)
// ...
```

or words to that effect (key point being to only create the parent)

faddat: I'll test with a gaia state sync...

boltdb.go
```diff
@@ -47,7 +46,11 @@ func NewBoltDBWithOpts(name string, dir string, opts *bbolt.Options) (DB, error)
 	}
 
 	dbPath := filepath.Join(dir, name+".db")
-	db, err := bbolt.Open(dbPath, os.ModePerm, opts)
+	if err := os.MkdirAll(filepath.Dir(dbPath), 0o0700); err != nil {
+		return nil, err
+	}
+	dblocation := filepath.Join(dbPath, "boltdb")
```
creachadair, line 52: Wait, isn't the `whatever.db` supposed to be the database itself? That's how it's used in the code, anyway. The database may be a directory in cases like Badger or LevelDB, but I thought the point of this exercise is that BoltDB is a flat file.

If we're just going to add another level of directory, we've not really addressed the original problem.

faddat: I think that I could put them all in data, then we could skip making directiies.

faddat:
PS: Can confirm that this and rocks now work for gaia.


Screen Shot 2022-05-07 at 2 27 56 AM

boltdb.go
faddat: cd boltdb
cd: not a directory: boltdb

faddat:
OH dear.  Works for gaia but does not pass tests.

boltdb.go
creachadair: > I think that I could put them all in data, then we could skip making directories.
> 
> This does not make another layer tho-- depends on the function that we want:

Right. I meant that it makes another layer in this particular case (`data/blah.db/boltdb` instead of `data/blah.db` being the actual Bolt file). But it seems to me like we want the path that the caller passes in to be what actually gets opened, and if we modify it here at creation time, they're actually getting a different result.

As you say, for databases that already create a directory we're getting that anyway, but everything to the right of that last slash in the dbDir path belongs to the backend, which seems like the right separation of concern.

faddat: Is this the only package that makes these folders?

I get a funny feeling about how gaia will start with this version that doesn't pass tests-- and won't start but tells me the folder is already there, otherwise.

creachadair: BoltDB may well be the only backend that uses a flat file right now.

As a matter of API design, I would like to have a simple, predictable result. One way to do that would be to to enforce a convention that the caller provides the "root" directory (e.g., the `$TMHOME/data` part in Tendermint parlance), and the backend is responsible for the rest of the path.

So if the caller passes in dir `$TMHOME/data` and name `something.db`, then the backend should create a database _at_ that path (`$TMHOME/data/something.db`) rather than adding additional subcomponents. 

Of course, the backend will put additional files inside that directory, but for the caller, what matters is what they would need to pass in to re-open the same thing later.

There are other ways to go about it, but that seems like maybe the simplest. And we're pretty close to that now, I think: The backends that create directories already seem to do this (or we could easily fix them to do so). If the BoltDB backend puts a plan file at the end of the path instead of a directory, that's fine! The caller doesn't care as long as the implementation handles its part of the contract.

I agree it'd be sensible to be consistent in some way, regardless.

PR #237, opened 2022-04-25
Omnibus update branch

faddat:
This branch is being used in production by notional on evmos.

The performance is strikingly better, see #248

This branch can be built with ease for chains with high resource demands as follows:

https://github.com/tendermint/tm-db/issues/248

```bash
docker run ghcr.io/faddat/cosmos -c /bin/bash
git clone your-chain-repo
cd your-chain-repo
go mod edit -replace github.com/tendermint/tm-db=github.com/notional-labs/tmdb@commit
go mod tidy
go install -ldflags '-w -s -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb' -tags rocksdb ./...
```

above you can change rocksdb to the database you'd like to try.

In total, this branch integrates

* a fix to rocksdb that allows it to make the snapshots folder
* migration from cosmos/gorockdb to a standard, well-maintained rocks wrapper, linxgnu/grocksdb
* Terra's removal of the prefixdb mutex
* Additional tweaks to goleveldb and rocksdb from the terra team

PR #250, opened 2022-05-07
feat(goleveldb,rocksdb): optimization: remove redundant check from iterator

faddat:
This is a cherry pick from terra's performance oriented branch.

It removes an unnecessary check when we use the iterator.


faddat:
@creachadair I really couldn't say for sure if this one is safe.

08d2:
> It removes an unnecessary check when we use the iterator.

How do you know it is unnecessary?

faddat:
@08d2 I don't for sure, which is why I said

```

@creachadair I really couldn't say for sure if this one is safe.

```

08d2:
> @08d2 I don't for sure, which is why I said
> 
> ```
> 
> @creachadair I really couldn't say for sure if this one is safe.
> ```

Then I suppose this PR should be closed, no?

faddat:
None of us should be sure.

My preference: I change this to a draft, and figure out if it's safe.

faddat:
@creachadair on this, I'd kindly request aid in figuring out how to make the certain determination that this code is safe.  It's part of the omnibus upgrade branch.  My comments about safety are driven by disabling these tests:

08d2:
@faddat You have removed a fair amount of code which verifies certain invariants during tests. If that code was added again, would the corresponding tests fail? If they would fail, can you explain why those failures are safe to ignore?

faddat:
Hi @08d2 -- I have mainly experiential info on this.  Basically, those iterator checks that are removed here, are what are required to pass the the tests that I removed here.

"it works".

Now, if you're able to help come up with some new tests, that would be super useful.  For both this and #237 -- the reality is that I'm very literally just doing what I can observe to work well. 

I would like to improve testing, and restructure this repo, and ensure that we remove badgerdb from future timelines for both the cosmos-sdk and tendermint -- or have very good reason for using it.

08d2:
But why did you remove those tests?

faddat:
ah-- look at the corresponding code

08d2:
Looking at the code explains the "what" but not the "why" — can you ELI5? 😇

faddat:
sure.   The tests were dependent on the "redundant check from iterator".

faddat:
https://github.com/tendermint/tm-db/pull/250/files#diff-61cc79a7f545ad48e7bce30b84f13aa4cfd4250b9898f0f6a18ccb38984acf10


See checknextpanics?

That would return the specific panic that the tests were looking for.

08d2:
I'm sorry, I don't follow. I'd appreciate it if you could take a few minutes to write out exactly what this PR is doing, and why you believe it is (or could be) safe. Connect the dots for me :) because at the moment, it's not obvious to me why any of these changes have been made, nor that any of them are actually safe.

faddat:
* https://github.com/tendermint/tm-db/pull/280

-- anyhow @08d2 -- I'd prefer to keep this as a draft, walk you through reasoning too. 

This PR, which is also contained in #237 -- is based on some work from Terra on performance, you can find it at their tm-db fork:

https://github.com/terra-money/tm-db

Unfortunately, they weren't super insistent on the tests passing there.  

1) it's possible that this PR is entirely in error eg: maybe their code wasn't safe, either.  
2) Seeking help to either:

* "this is plain usafe, close"
* make better tests

PR #251, opened 2022-05-07
gitpod

faddat:
Currently it is a little tough to work on this code.

The gitpod environment here provides an image that has all needed dependencies. 

This PR depends on 

#236 to work properly.


creachadair:
I don't have a good sense of what problem this is solving. Generally I'd expect changes to tm-db itself should be fine to build and test locally.  Am I missing some other concern?

Regardless: If we're going to add new dependencies on container images somewhere, ideally we should pick ones that are built from the same repo. (I don't think you're planning anything nefarious, but it's harder to audit and fix issues in images built and hosted externally).

faddat:
I definitely understand. 

I'm trying to get that image builder hosted in the cosmos org, and I'll keep you up to date.  Here's what we can do with it:

https://github.com/notional-labs/tm-db/runs/6345821218?check_suite_focus=true

faddat:
Ah, so gitpod isn't a workflow system, it is a code editor. 

I use it so that I don't need to check out repositories onto my computer, and need to worry about the local state of my git repositories.

faddat:
@creachadair so I just changed the docker image.  

Now I will use gitpod to solve the remaining issues in #239 and I will be able to run it in exactly the same environment as CI, without a lot of additional work.

Issue #186, opened 2021-09-03
memdb: Implement btree.Item on `item` directly, in addition to `*item`


ValarDragon:
Currently we force every btree item to be `*item`, not just `item`. This in turn causes every set to require new heap allocations, which turns out to be a significant overhead. These allocations cause lots of memory to need to be garbage collected, and this ends up often being the bottleneck, not the actual btree searching operations.

In Osmosis, we get that we have ~no item spent in the btree operations, its all spent in allocating the `newPair` onto Heap. Heres a screenshot of a relevant flamegraph output here: 
<img width="1502" alt="Screenshot 2021-09-03 at 12 28 35 PM" src="https://user-images.githubusercontent.com/6440154/132045005-770a203e-d22f-4c0c-b6eb-d9a8ba03cbeb.png">

marbar3778:
Hey, I can help in getting a release out to fix this, if you are willing to make a PR. I can push a branch for you to target as master has diverged too much to be used for this

ValarDragon:
smh stale bot

ValarDragon:
Can we please remove stale bot issue closing =/ Its not even that long of a wait atm

marbar3778:
yea ill fix

faddat:
What's the status of this one?

I guess I want to include it in the notinal tracking issue

@ValarDragon do you think this is needed still?

ValarDragon:
Yeah this is still useful

faddat:
great!

faddat:
So, fair to say this one is more important than #188 ?

that’s good because ah, wow

Pain is a teacher I guess?

faddat:
Seems set :)

PR #254, opened 2022-05-07
implement btree.item on item itelf, not only its pointer.

faddat:
Closes: #186 

@ValarDragon I think that this is correct, but I did have one question

### DIFF
--- a/memdb.go
+++ b/memdb.go
@@ -32,6 +32,14 @@ func (i *item) Less(other btree.Item) bool {
 	return bytes.Compare(i.key, other.(*item).key) == -1
 }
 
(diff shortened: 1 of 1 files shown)


memdb.go
```diff
@@ -32,6 +32,14 @@ func (i *item) Less(other btree.Item) bool {
 	return bytes.Compare(i.key, other.(*item).key) == -1
 }
 
+// LessDirect implements btree.Item directly on item instead of its pointer.
+// Question: is this it, or do we need to get the pointer off of other.(*item).key?
```
faddat, line 36: this is the question

faddat:
Thanks @catShaark !

creachadair (approved):

PR #258, opened 2022-05-19
Use ghcr.io and build docker for arm, too

faddat:
This PR moves the docker images away from docker hub, to the github container registry, making them more visible.

Additionally for greater ease of use for devs on M1 Macs, this PR also makes a multiplatform docker container.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -25,7 +25,7 @@ jobs:
       - name: Prepare
         id: prep
         run: |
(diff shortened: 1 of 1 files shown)


faddat:
We're currently underestimating the level and depth of gymnastics needed to work on tm-db in an environment with all of the databases supported.  

To address this, I am also adding a .gitpod.yml file to this repository that corresponds to the docker image made in ci.  Really.  It is really hard and eats time.

PR #274, opened 2022-07-21
Standardized Development Environment

faddat:
## Description

Yesterday I made some changes to the standard dev env that we're using and to my surprise, they "just worked".  That includes adding docker to the image so that we could do protocol buffer builds.  It also has all the rocksdb tooling installed in the filesystem for those with container allergies. 

Caveat: the rocksdb installed here is of the v7 variety.  We would want to update our rocks library to grocksdb.

```yaml
image: ghcr.io/notional-labs/cosmos
```

That image is built here daily using cron via github actions and each build is observable. 

https://github.com/notional-labs/containers/blob/master/cosmos/Dockerfile


.gitpod.yml
```diff
@@ -0,0 +1 @@
+image: ghcr.io/notional-labs/cosmos
```
marbar3778, line 1: what's wrong with the Tendermint-dev docker image?

faddat: I should close this PR, because if we change to the notional one it should be done as a part of #278 

in 278 we would become incompatible with the tendermint-dev docker image.

PR #276, opened 2022-07-21
remove badgerdb and boltdb

faddat:
This PR attempts to lower the matienance scope of the tm-db repository by removing databases with no known users.

Further, badgerdb has now entered the status of abandonware and is no longer actively maintained.


marbar3778:
I love the reduction. Could you add a changelog entry?

I guess the issue here is if someone is using it they will need to resync the db. Maybe we mark them as deprecated and put a EOL date a few months into the future?

faddat:
I actually don't think it is needed (the EOL) because they could always bump the version downwards, and maybe more importantly, I don't know of any successful deployments of badger or bolt.

None of the changes here break the api, so the bumping down technique should work quite fine

Karam19925 (approved):
#276

Karam19925 (approved):
#276

Karam19925 (approved):

Karam19925 (approved):
> #276 checked

Karam19925 (approved):
checked

marbar3778:
@mvdan I believe you added badger, do you use it?

mvdan:
Not anymore for new projects; it's been semi-abandoned for some time and it's hard to run smoothly at scale. I've rediscovered sqlite in the meantime :)

faddat:
Hi @mvdan -- just wanted to say thanks for gofumpt!

I'm a daily user :)

Karam19925 (approved):
> #276

Karam19925 (changes requested):
> #276 remove

PR #278, opened 2022-07-21
Grocksdb and docker

faddat:
This PR introduces grocksdb and a docker container that has rocksdb v7.x in it


.github/workflows/ci.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     # rocksdb, which are needed to build and test those modules.
     # To update the container image, see docker.yml.
     runs-on: ubuntu-latest
-    container: tendermintdev/docker-tm-db-testing
+    container: ghcr.io/faddat/cosmos
```
marbar3778, line 21: reasoning?

.github/workflows/lint.yml
```diff
@@ -1,19 +1,42 @@
...
+  # Optional: allow read access to pull request. Use with `only-new-issues` option.
+  # pull-requests: read
 jobs:
   golangci:
-    # We need to run the linter on the same image we use for building, since it
-    # needs the C libraries installed for the dependencies to typecheck.
+    name: lint
     runs-on: ubuntu-latest
-    container: tendermintdev/docker-tm-db-testing
+    container: ghcr.io/faddat/cosmos
```
marbar3778, line 13: ditto

.github/workflows/lint.yml
```diff
@@ -1,19 +1,42 @@
...
+
+          # Optional: if set to true then the all caching functionality will be complete disabled,
+          #           takes precedence over all other caching options.
+          # skip-cache: true
+
+          # Optional: if set to true then the action don't cache or restore ~/go/pkg.
+          # skip-pkg-cache: true
+
+          # Optional: if set to true then the action don't cache or restore ~/.cache/go-build.
+          # skip-build-cache: true
```
marbar3778, line 42:
```suggestion
```

marbar3778 (commented):
Could you add a changelog entry and some reasoning as to why this change is needed

faddat:
You're right it definitely needs docs.

.github/workflows/ci.yml
faddat: reasoning:  when we bump rocksdb, we also need to use a different docker image.  The alternative is to upgrade the current docker image.

.github/workflows/lint.yml
faddat: reasoning: when we bump rocksdb, we also need to use a different docker image. The alternative is to upgrade the current docker image.

PR #279, opened 2022-07-26
no pedantic

### LINKED ISSUE #263: rocksdb: pedantic error checks that could just be direct returns
If we examine the code in RocksDB, we can see this pedantic pattern that could instead be a direct return
image

faddat:
Thanks!

Closes #263


marbar3778 (approved):
thanks for the cleanup

marbar3778:
@faddat could you update to latest master and then we can merge

Karam19925 (approved):
> #279 checked

Issue #275, opened 2022-07-21
TestRemoteDB fails in main branch


faddat:
* #274 shows this.

faddat:
```
ok  	github.com/tendermint/tm-db	1.306s	coverage: 80.8% of statements
=== RUN   TestRemoteDB
    remotedb_test.go:32: 
        	Error Trace:	/__w/tm-db/tm-db/remotedb/remotedb_test.go:32
        	Error:      	Expected nil, but got: &status.Error{s:(*status.Status)(0xc000186048)}
        	Test:       	TestRemoteDB
--- FAIL: TestRemoteDB (0.02s)
FAIL
coverage: 4.8% of statements
FAIL	github.com/tendermint/tm-db/remotedb	0.053s
testing: warning: no tests to run
PASS
coverage: 0.0% of statements
ok  	github.com/tendermint/tm-db/remotedb/grpcdb	0.045s	coverage: 0.0% of statements [no tests to run]
```

faddat:
The certificates in the remotedb folder had expired.

PR #280, opened 2022-07-26
regenerate certs

faddat:
I ** think** that these certificates had expired, and this was causing #275.

If this PR passes tests here (it passes locally for me) then this is the fix.


faddat:
Note:  This one really should be merged.

Master is broken.

https://github.com/tendermint/tm-db/runs/7574887469?check_suite_focus=true

marbar3778 (approved):

PR #283, opened 2022-07-28
Pebble

faddat:
Pebble is a rocks-like kv store written in pure go.

* no cgo
* faster than rocks especially when used with cw
* faster than goleveldb
* more meaningfully configurable than goleveldb
* complies with all tm-db tests, and tested with additional scrutiny by Notional

pebble is the best possible candidate for "the universal cosmos kv store".


faddat:
wrong branch, sorry if anyone used this one.

PR #284, opened 2022-07-29
Pebble

faddat:
This branch includes work from myself, @catShaark, and @baabeetaa.  @baabeetaa solved the final issues in the pebble implementation for tm-db.

It's been running on Sifchain for about 15 hours.

We will soon have pebble snapshots for every chain that Notional validates available at:

https://github.com/notionla-labs/cosmosia

But we didn't just test this against Sif, that has led to failures in the past, so instead, we tested it millions, and millions of times, against both the tm-db test suite and iavl. 

Like the other databases, pebble would perform faster if this pr:

* https://github.com/tendermint/tm-db/pull/239

were merged.

Like all the PR's here, this pr would pass tests if this pr:

* https://github.com/tendermint/tm-db/pull/280

were merged -- it fixes the master branch of this repository.


faddat:
Passes tests because merged #284

db.go
```diff
@@ -28,13 +28,19 @@ const (
...
 	//   - use rocksdb build tag (go build -tags rocksdb)
 	RocksDBBackend BackendType = "rocksdb"
-
+	// BadgerDBBackend represents badgerdb (uses github.com/dgraph-io/badger)
+	//   - EXPERIMENTAL
+	//   - use badgerdb build tag (go build -tags badgerdb)
 	BadgerDBBackend BackendType = "badgerdb"
+	// PebbleDBDBBackend represents pebble (uses github.com/cockroachdb/pebble)
+	//   - EXPERIMENTAL
+	//   - use pebble build tag (go build -tags pebbledb)
```
marbar3778, line 42: I believe this is pure go so no need for a build flag.

faddat: We tag all of them currently (badger, bolt, and pebble)

faddat: I think that if we don't tag, then we end up making chain binaries that use more than one db.  This could (or might not) be desirable....

faddat: if we want to make it default, then no flag.  But I think default is like maybe next month.

Cause it is 10-27x less intense on the disk, no joke.

faddat: https://twitter.com/gadikian/status/1553351762894917632?s=20&t=eXmh9UbBSYDnBO3kRFbh2w

sascha1337: 27x aint a joke, true

faddat: eh, need to do more benchmarking.  its.... wierd. 

But:

* near certain drives will last longer
* near certain will be faster with cw
* near certain will enable running on much cheaper nodes

faddat: I took the tags off.  Not sure what'll happen with tests and such

itsdevbear:
@faddat how has pebbledb performance been compared to rocks? I am very much ready to ditch CGO. Any benchmarks yet?

sascha1337:
> @faddat how has pebbledb performance been compared to rocks? I am very much ready to ditch CGO. Any benchmarks yet? 

Check notional org repos, cosmodia was it I think, they testing it in prod. right now on 40 chains or something

PR #297, opened 2022-09-29
support mdbx backend

yihuang:
I added this backend to run some benchmarks, to see how mdbx performs.
Based on the simple local benchmarks, compared with rocksdb, the db size is larger, insert speed is slower. maybe interesting to try on a production node.

- The batch might can be implemented more efficiently using the transaction, if we only use one batch at the same time.
- It don't use the other features like dupsort.

One feature that would be very useful is the db can be opened by multi-processes, so one can inspect the db of a live node.

### DIFF
--- a/mdbx.go
+++ b/mdbx.go
@@ -1,0 +1,193 @@
+//go:build mdbx
+// +build mdbx
+
+package db
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/c2h5oh/datasize"
+	"github.com/torquem-ch/mdbx-go/mdbx"
+)
+
+func init() {
+	registerDBCreator(MDBXBackend, NewMDBX, false)
+}
+
+type MDBX struct {
+	env *mdbx.Env
+	dbi mdbx.DBI
+}
+
+var _ DB = (*MDBX)(nil)
+
+func NewMDBX(name string, dir string) (DB, error) {
+	path := filepath.Join(dir, name) + ".db"
+	env, err := mdbx.NewEnv()
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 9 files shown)


mdbx.go
```diff
@@ -0,0 +1,193 @@
...
+		return errValueNil
+	}
+
+	return db.env.Update(func(txn *mdbx.Txn) error {
+		return txn.Put(db.dbi, key, value, 0)
+	})
+}
+
+func (db *MDBX) SetSync(key []byte, value []byte) error {
+	// TODO
```
yihuang, line 115:
```suggestion
```

faddat:
Sir, I will review and test, no matter tm-db total lack of matienance.

I had no idea how to implement.  thanks!

tac0turtle (commented):
tendermint should use a single kv db and get rid of the rest or migrate to sqldb. Adding dbs without proper testing and optimisation for the use case is adding more noise.

yihuang:
> tendermint should use a single kv db and get rid of the rest or migrate to sqldb. Adding dbs without proper testing and optimisation for the use case is adding more noise.

I was actually trying this backend with cosmos-sdk, the iavl tree store, agree that rdbms could fit tendermint well.
Since cosmos-sdk is migrating to another repo, maybe we leave it as an experimental thing, no need to merge.

faddat:
Will try to add this to cosmos-db instead.  

This PR:

https://github.com/cosmos/cosmos-sdk/pull/13966

and /or this one:

https://github.com/cosmos/cosmos-sdk/pull/13931

should help, I thought they were great ideas.

faddat:
https://github.com/tendermint/tendermint/pull/9742

tac0turtle:
not sure how sdk store package PR helps this PR? could you elaborate?

PR #310, opened 2022-12-18
support create custom rocksdb db instance

yihuang:
- for easier tuning rocksdb options.
- support open in read-only mode
  - useful in some offline commands
  - could be used in grpc-only mode, rocksdb support multi-process open in read-only mode.


yihuang:
unstale

